PR #133, opened 2026-07-18
fix(directory): officials cards — single action row, full-width content

ashokhein:
Per request: the per-card button stack moved into one row.

Before: a vertical right rail (status chip + Invite + Edit roles + Delete) squeezed the name column — noticeably cramped on mobile.
After: top-right keeps only the Linked/Invited chip; all actions in a single wrapped row under the content, Delete right-aligned. Verified at 1280px and 390px including the expanded invite form (no overlap).

Gates: tsc 0 · new markup test 3/0.


ashokhein:
Desktop / mobile / expanded-invite screenshots from the local verification run:

PR #184, opened 2026-07-20
fix(ui): save-point button, danger zone, Constraints rules sheet, and one home for exports

ashokhein:
A run of console UI defects found by reading the screen rather than the code.

### Save point button was "fat"
`.btn` has no `whitespace-nowrap` and no `shrink-0`; the sibling input is `w-full`. Flex squeezed the button until "Save point" **wrapped onto two lines** — a 60px solid-purple slab beside a 40px input.

### "Clear schedule…" wasn't a button
Styled `btn border-red-200 text-red-700`. `.btn` carries padding but no border *width* and no background, and `border-red-200` is a colour with nothing to paint — so the most destructive control in the panel rendered as bare red text, its padding showing only as a stray indent. `.btn-danger` is the existing token for this. Swept the app: the two other `btn` + border-colour call sites both ride `btn-ghost`, which supplies the width.

### Constraints tab
Rebuilt as a rules sheet — each rule stated in plain English on the left, its control right-aligned in one scannable column, rows stacking below `sm`. The old layout put a `w-full` select inside a flex label, so "Field fairness" wrapped and the select swelled to fill the card.

**Field fairness got a help chip**, because nothing on screen said what it did. It is a tie-break, not a rule: among courts free at the same instant, prefer the one an entrant has used least (`balance`) or did not just play on (`rotate`). It never delays a match. Copy in all four locales; the `/help` scheduling article gains a section.

The chip sits *outside* the `<label>` deliberately — `Tip` renders a `<button>`, and a button inside a label both pollutes the accessible name and re-targets clicks at the select.

### Exports had two homes
Five loose buttons in the schedule header, plus a Documents menu on the fixtures view. Now one — the menu, which gains **Rosters, Standings and Participants** so nothing is lost. That needed two new row shapes: XLSX-only (participants has no print edition) and extra query (standings prints landscape). URL building moved into one helper so the busy-state check can't drift from the URL actually requested.

The menu was also a bare `<details>`, which only ever toggles from its own `<summary>` — it stayed open over the fixtures list until you travelled back to the trigger. Now closes on outside pointerdown or Escape.

### Also
Court name inputs no longer span the full card.

---

**Behaviour change worth a look:** `Match sheets` no longer passes `pageBreaks=per_pitch`. The old schedule-page button did; the menu row never has. I kept the menu's existing behaviour rather than silently changing it for everyone — say the word if per-pitch breaks were the point.

**Verified:** tsc clean, eslint clean, 890 tests across `src/components` and `src/lib`. Every fix ships a test that fails without it (proven by stashing). Screenshotted at 1440px and 390px.


ashokhein:

PR #185, opened 2026-07-20
fix(engine): make the placer and the verifier agree about constraints

ashokhein:
There are two engines. `slotFixtures` **places** fixtures; `validateAssignments` **verifies** them. The placer honoured the full constraint set; the verifier implemented a subset — and the verifier is what guards the board and the AI referee.

So the same timetable was legal or illegal depending on which door it came through.

### Rest was resolved three different ways

Rest is stored twice — `perEntrantMinRest` (Settings, the shape of the day) and `constraints.restMin` (Constraints, a rule about entrants):

| path | rest used | |
|---|---|---|
| `slotFixtures` | `max(both, restByGroup, noBackToBack)` | correct |
| board validation | `perEntrantMinRest` only | ignored `restMin` |
| AI referee | `constraints.restMin` only | ignored the other |

Mirror-image bugs, one cause. Set rest in Settings alone and **AI Schedule silently ignored it** while Auto-schedule enforced it; set it in Constraints alone and the board ignored it.

`effectiveRestMinutes` is now exported from the engine and used by both, so they cannot drift. `noBackToBack` and `restByGroup` come along free — the verifier never checked either.

The board needed **no change**: it already passes the whole `SlotConfig`, so widening the parameter type was enough. The AI pack did — it had no rest field at all, which is exactly how the referee ended up reading only one of the two.

### Start windows were placement-only

A hard bound in the solver, never read by the verifier. So the bound held when Auto-schedule placed a match and evaporated the moment somebody dragged one. Bounds the START, matching the solver, and takes the tightest of every window targeting the fixture (entrant, pool or division).

### crossPersonClash="hard" wasn't honoured

The organiser asked for a person double-booking to be *refused*, not badged. The solver refuses to place one; the board accepted a hand-placed clash, because `mapConflicts` decided blocking from the reason alone and never consulted the setting. Default is `warn`, so only orgs that opted in are affected.

### Not in this PR

`parallelism: "block"` (a division holding exclusive slots against siblings). It needs a `ConflictReason` the taxonomy doesn't have, pulling in a `REASON_CODE` entry and client labels in four locales — its own change.

Longer term the two rest fields should collapse into one, but only once the engine agrees, not before.

---

**Rollout note:** this surfaces violations that currently pass silently. That's correct but will read as a regression on existing schedules. Start windows report as a badge; only `crossPersonClash: "hard"` newly *blocks*, and only where it was explicitly set.

**Verified:** tsc clean; 108 engine tests, 470 web server tests. 16 new tests, 8 of which fail without the change (proven by stashing). The one `key-scopes.test.ts` failure fails identically on clean `main` — it's what #183 fixes.

### DIFF
--- a/apps/web/src/server/usecases/__tests__/__snapshots__/schedule-ai-pack.test.ts.snap
+++ b/apps/web/src/server/usecases/__tests__/__snapshots__/schedule-ai-pack.test.ts.snap
@@ -765,6 +765,7 @@ exports[`buildSchedulePack (v4/01 §2) > pack is deterministic and matches the 0
     ],
     "gapMinutes": 0,
     "matchMinutes": 30,
+    "perEntrantMinRest": 20,
     "sessionWindows": [
       {
         "from": "2026-08-01T10:00:00+01:00",
--- a/apps/web/src/server/usecases/__tests__/schedule-ai-effort-ab.live.test.ts
+++ b/apps/web/src/server/usecases/__tests__/schedule-ai-effort-ab.live.test.ts
@@ -86,6 +86,7 @@ function teamsPack(): { pack: SchedulePack; movable: Set<string> } {
     settings: {
       matchMinutes: 30,
       gapMinutes: 10,
+      perEntrantMinRest: 0,
       courts: ["Court 1", "Court 2", "Court 3"],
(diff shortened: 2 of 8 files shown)


ashokhein:

PR #186, opened 2026-07-20
fix(exports): unoverlap the footer QR, and point it at the public site

ashokhein:
Two defects in the same PDF footer.

### The QR printed through the wordmark

The live-page QR is drawn at `(page.width - MARGIN - 28, fy - 22)`; the "Powered by seazn.club" attribution is right-aligned to the same edge at `fy`. Measured: the attribution's right edge landed at **555.28**, the QR started at **527.28** — a 28pt collision on every export with a public live page.

The attribution now yields 34pt (28pt code + 6pt gutter) when a QR is present, and keeps the full width when it isn't. Beyond looking wrong, an obscured finder pattern is a QR phones refuse to scan.

### The QR pointed at the wrong host

URLs were built from the request origin (`url.origin`), so a PDF generated in dev encoded `http://localhost:3000`, and one generated behind a proxy encodes whatever host the proxy presents. **A QR is scanned by a phone that has to reach the public site, so the request can never be the authority.**

Both the division live-page link and the admit-ticket `/r/<ref>` link now use `siteOrigin()` — the same `OAUTH_BASE_URL` / `NEXT_PUBLIC_BASE_URL` resolution the emails and canonicals already trust.

`origin` leaves `ExportOpts` and three route call sites entirely; one (`buildMyRotaDoc`) had been carrying it as `void origin` for "signature parity" without ever using it.

---

**Contract change:** `liveUrl` no longer depends on a caller threading an origin, so visibility alone decides whether a QR is drawn. Private competitions still get none — their `/shared` page 404s. The exports test asserting "no origin, no QR" is updated to match.

**Verified:** tsc clean, eslint clean, 55 server tests. The overlap test fails without the fix (proven by stashing) — the stock doc-render spy discards geometry, so this one records x/y/width.

### DIFF
--- a/apps/web/src/app/api/v1/competitions/[id]/exports/tickets/route.ts
+++ b/apps/web/src/app/api/v1/competitions/[id]/exports/tickets/route.ts
@@ -18,9 +18,7 @@ export async function GET(req: Request, { params }: Ctx) {
     const url = new URL(req.url);
     const format = url.searchParams.get("format") ?? "pdf";
     if (format !== "pdf") throw new HttpError(400, "format must be pdf");
-    const model = await buildAdmitTicketsDoc(
-      auth, id, { printedAt: new Date().toISOString() }, url.origin,
-    );
+    const model = await buildAdmitTicketsDoc(auth, id, { printedAt: new Date().toISOString() });
     const bytes = await docModelToPdf(model);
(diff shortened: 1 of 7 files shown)


ashokhein:

PR #187, opened 2026-07-20
chore(schedule): retire the flexible scheduling mode

ashokhein:
`scheduling_mode` was never selectable.

It is absent from the division creation flow, absent from the Settings tab, and **no component in the app reads the column**. The only way to reach `'flexible'` was to hand-patch the v1 API.

What an organiser got if they managed it:

| action | behaviour |
|---|---|
| Auto-schedule | 422, does nothing |
| Re-flow remaining | 422, does nothing |
| AI Schedule | 409, does nothing |
| The board | **ignored the mode entirely** — hand-build a timetable anyway |

Three dead buttons and one inconsistency, guarding a mode nobody could choose.

### Removed
- the auto-schedule guard (`schedule.ts`) and the AI pack guard (`schedule-ai.ts`) — those paths now treat every division alike
- `scheduling_mode` from the AI pack, where it was sent to the model as metadata that no longer means anything
- `scheduling_mode` from `PatchDivision`, so nothing can set it again

### Kept
**The column stays, dormant**, until a later migration drops it. Guards are revertible from git; a dropped column with data is not. It remains on `DivisionOut` for the same reason.

This also closes a bug for free: the board never checked the mode, so retiring it makes that inconsistency disappear rather than needing its own fix.

---

**Note on the spec diff:** the regen also picks up pre-existing drift — checkpoint `kind` (`manual|ai`) from #181, which merged without a regen because the openapi gate is path-filtered and skips on `main`. Correct to include, but not part of this change.

**Data:** local dev shows 138 flexible divisions / 557 fixtures / **0 with times**, all of which look like test artifacts. Production was not checked — the user chose to proceed without it.

**Verified:** tsc clean; 470 server tests. The `key-scopes.test.ts` failure fails identically on clean `main` — it's what #183 fixes.

### DIFF
--- a/apps/web/src/server/api-v1/schemas.ts
+++ b/apps/web/src/server/api-v1/schemas.ts
@@ -129,8 +129,6 @@ export const PatchDivision = z
     status: DivisionStatus,
     /** Hide official names on all public reads (Jul3/02, 25 Jun). */
     officials_hide_names: z.boolean(),
-    /** Jul3/04 §4: 'flexible' = ordered fixtures, no clock. */
-    scheduling_mode: z.enum(["timed", "flexible"]),
     /** Jul3/08 §5: progression fires without a button (Pro formats.advanced). */
     auto_progress: z.boolean(),
     /** SPEC-2: draft a news post when results land (Pro `news.auto` on write). */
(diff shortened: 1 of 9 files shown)


ashokhein:

PR #188, opened 2026-07-20
fix(wizard): stop the division wizard silently dropping its scheduling step

ashokhein:
The division wizard's Scheduling step collects courts, match length, start date and end date, then seeds them with `PUT /divisions/:id/schedule-settings` right after create.

That PUT is wrapped in a bare `catch { /* board settings are editable later — ignore */ }`.

And `usesConstraints()` (`schedule.ts:81`) returns true when **`config.courts.length > 1`** — a second court is a Pro feature.

So a Community organiser who adds a second venue in the wizard gets a 402 on the whole settings PUT, the catch swallows it, and they lose **match length, start date and end date too** — none of which had anything to do with the Pro gate. Silently, with no error shown.

### Fixed
- venue list gated behind `constraintsAllowed` — an `UpgradeGate` and a hint instead of an "Add venue" button that leads to a silent 402
- the seed retries with a single venue if the PUT fails, so the rest of the step survives
- fallback venue uses the sport's own noun rather than a hardcoded `"Court 1"`
- help pages corrected: they claimed multi-court was available on all plans, and neither mentioned that the wizard's Scheduling step sets these

### Scope note
The original brief asked to *move* these fields from Settings into the wizard. They were already in the wizard — and `board/settings-panel.tsx` is the only post-create editor for them, so removing them there would have left no way to move a rained-off start date or rename a venue. Not done, deliberately.

Play hours deliberately stay out of the wizard: empty means all hours playable, and the answer is better made on the board with the timetable in front of you than guessed before any fixtures exist.

**Wizard field count: 4 before, 4 after.**

---

Spotted, not fixed: the venues label pluralises naively — football renders "Pitchs".

**Verified:** tsc clean; 551 tests pass across `src/lib` and `src/components/v2`. 6 new tests; **5 fail without the fix**.

### DIFF
--- a/apps/web/content/help/getting-started/generate-fixtures.md
+++ b/apps/web/content/help/getting-started/generate-fixtures.md
@@ -12,7 +12,7 @@ Generating is **idempotent**: pressing it again after adding two late entrants o
 
 ## Times and courts
 
-Generated fixtures have an order but no clock until you schedule them. Use the **schedule board** to drag matches onto courts and times, or **Auto-schedule** to lay the whole day out in one pass, then fine-tune. Play hours, court counts and rest gaps live in the division's schedule settings. More in [The schedule board](/help/scheduling/board).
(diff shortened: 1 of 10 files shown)


ashokhein:

PR #460, opened 2026-08-04
W4a follow-ups T2+T4: an additive tripwire for state/cfg, and lint coverage for packages/engine

ashokhein:
Second half of the W4a audit follow-ups: **T2** (state/cfg additive tripwire) and **T4** (lint coverage for `packages/engine`). Both are test/infra-only — no engine behaviour changes, and the frozen `*.golden.json` corpora are byte-identical apart from T2's sanctioned `EXTEND_GOLDEN` appends.

First half is #454 (T1 + T3).

## T2 — the tripwire the goldens didn't have

`state` and `cfg` had no additive tripwire, so a *narrowing* schema change could land green.

The design changed during the work, for a good reason. The original brief proposed freezing an explicit set of state paths — but `stateMismatch` already compares everything outside `cfg` as **exact string equality**, so a frozen path-set restates that more weakly, and cricket's set is polluted by 22 person-keyed record paths. The real gap was never the comparison, it was **coverage**: a field no stream exercises is a field no corpus can defend. So the declared side now comes from a seeded sweep of each module's own `arbitraryEvent`, diffed against the frozen corpora.

- All nine previously-unpinned fields are now covered; the unreachable list ships **empty**.
- Four were already covered and are now pinned. Five needed new coverage — including tennis `overran`/`overCount`, where **no variant declared an allowance**, so no fold had ever computed them.
- Fixed a bug in the existing walker: `fieldPresent` never descended a `z.record` for `[]`, so `suspensions.classes[].pim` read as unpinned while the frozen cfg pinned it.

Golden appends are additive and were verified against `main` independently of the agent that wrote them: boardgame 74→75 (`timed`), cricket 25→26 (`reviewed`), football 23→24 (`lawful`), hockey 19→20 (`fih-detail`), icehockey 13→14 (`iihf-detail`), tennis 21→22 (`breaks`). Prefixes byte-identical, no pre-existing config entry mutated, no scalar drift.

**Red-first evidence.** Narrowing `football.subWindows` gives 118/4 red on the new corpus — but only 120/2 on the *pre-change* corpus, and those two reds are the new gate itself. Every replay and parse test stayed green before this work: the mutation was **invisible**. That is the gap, demonstrated rather than asserted.

## T4 — `packages/engine` had no lint coverage at all

New `packages/engine/eslint.config.mjs` (`@eslint/js` + `typescript-eslint` `recommendedTypeChecked`), a `lint` script on the package, and `113 problems → 0`.

Deliberately **not** a copy of `apps/web`'s config — that one is `eslint-config-next` (React/JSX/hooks/router), meaningless for a dependency-light library. Type-aware linting costs ~16s and is what makes `no-floating-promises` / `await-thenable` work across the async z3 bridge and `no-unnecessary-type-assertion` work on the kernels.

Commits are split so the diff is readable: config → 101 autofixes (redundant `as`/`!`, type-erasure only) → four judgemental fixes → suppressions → override narrowing.

**No rule is disabled globally.** `reportUnusedDisableDirectives: "error"` proves every one of the 18 remaining suppressions is load-bearing.

Two things worth knowing for anyone touching this later:

- `no-unnecessary-type-assertion`'s **autofix is unsafe in this package**. It ignores narrowing and judges against the assignment target's declared type; it broke `tsc` with 10 errors at 3 sites (`open = openInnings(next) as T` in cricket ×2, setbased ×1). Those carry per-site disables. **Any future `--fix` here must be followed by `tsc`.**
- `allowDefaultProject` rejects `**` outright (`validateDefaultProjectForFilesGlob.js:20`, a deliberate perf guard), so `scripts/*.ts` is the only legal form even though the override next to it is `scripts/**/*.ts`. The asymmetry is documented in-file.

### Review round

The reviewer confirmed behaviour was untouched (every removed `as` is in expression position; no `!` removed anywhere; `officials/assign.ts:294` kept its `.fill(0)`, so the array stays dense) and caught two overrides that were far wider than the seams they named. Both were fixed in `dbcc580a`:

| override | before | after |
| --- | --- | --- |
| testkit | `src/testkit/**/*.ts` — ~3.8k lines incl. the **public `./testkit` entrypoint** | `schema-fields.ts` + `golden.ts` only |
| z3 bridge | `z3-*.ts` **+ `repair.ts`** (773-line orchestrator, not the bridge) | `z3-*.ts` only |

Narrowing the first surfaced 12 more findings, all fixed at the site rather than by widening the pattern back — 7 traced to a single root cause (`AnySportModule = SportModule<any,any,any>`), fixed by pinning `: unknown` at the 5 boundaries where the `any` enters.

Also dropped `measureLoadFactor()` from `repair-decompose.test.ts`, which burned ~100ms per run for a value nothing consumed. The flat `SOLVE_TIMEOUT` there is correct as-is — it is only vitest's bail-out ceiling, and the load-sensitive assertions already go through `timedUnderLoad`.

## Verification

Every number below was re-run by me on a clean, quiescent tree from the worktree, not taken from an agent's report.

| gate | result |
| --- | --- |
| `packages/engine` vitest | **2152 passed / 0 failed / 2153 total** (1 pending), 558 suites, 0 failed suites |
| `packages/engine` eslint | exit 0, **zero bytes** of output |
| `packages/engine` tsc | exit 0, zero output |
| `apps/web` tsc | exit 0, zero output |
| `apps/web` testkit consumer (`scoring-vocab`) | 17/17 |
| goldens vs `main` | T2 appends only; **nothing** touched by T4 |
| openapi drift | none |
| CI actually lints the engine | `turbo run lint --dry=json` resolves `@seazn/engine#lint` + `@seazn/web#lint`; `ci.yml:122` runs it |

`apps/web` lint is unchanged at its baseline `79 problems (0 errors, 79 warnings)`.

Baseline note: the `2152` here is measured against `main`, which is 22 commits behind `fix/w4a-followups` — so it is not comparable to the `2170` quoted on #454. T2's own delta is **+78 tests**.


ashokhein:
Rebased onto `main` after #454 merged. Conflicts were the five golden corpora — resolved the sanctioned way rather than by hand-merging frozen state strings: take `main`'s corpus verbatim, then re-run `EXTEND_GOLDEN=1`. Re-verified additive against `origin/main`, existing streams byte-identical, no pre-existing config entry mutated:

| corpus | streams | new config |
| --- | --- | --- |
| cricket | 26 → 27 | `reviewed` |
| football | 24 → 25 | `lawful` |
| hockey | 20 → 21 | `fih-detail` |
| icehockey | 14 → 15 | `iihf-detail` |
| tennis | 22 → 23 | `breaks` |

Counts start one higher than the original push because `main` itself gained a stream from #456/#457 — this is a re-append onto the newer corpus, not a double-append. The greedy scan picked the same five stream names unprompted.

**The rebase also produced the first real evidence that T4's gate earns its keep.** Engine code merged from #454/#456/#457 predates the package having any lint config, and it arrived carrying 10 `no-unnecessary-type-assertion` errors (`f458c851`). Nine were type erasure. The tenth was the exact trap this config's comments warn about, hit in code the original pass never saw: at `nested/kernel.ts` the rule judged `as "deuce" | "ad"` against the ternary alone and called it redundant, but the literals widen to `string` through the spread below, so the autofix broke `tsc` with a seven-line type error. Fixed by annotating the binding — satisfies rule and compiler, leaves no suppression behind.

Re-verified on the rebased tree: engine **2276 passed / 0 failed / 2277 total**, 598 suites, 0 failed suites; `eslint` exit 0 with zero output; engine and `apps/web` `tsc --noEmit` both exit 0 zero output; working tree clean.

One transient red worth recording rather than hiding: an interim run failed `repair-scale` "500-movable board returns inside its budget" at 9090ms against a 7000ms ceiling. That was a wall-clock assertion losing to eslint and tsc running concurrently on the same machine; it passes on a quiet one. Not diff-related, but it is load-sensitive in CI too.

PR #480, opened 2026-08-05
TypeScript 7 + Node 26 + pnpm: a ~7.5x typecheck and a ~3x worktree install, with no behaviour change

ashokhein:
Moves the toolchain to **TypeScript 7 + Node 26 + pnpm**, as three stages built and gated separately on a long-lived `typescript7` branch, then raised to `main` once.

No application behaviour changes. The evidence for that claim is that e2e came back **301 passed / 25 skipped / 0 failed at every stage** — identical run for run.

## Headline

| | before | after |
| --- | --- | --- |
| `tsc` (apps/web) | 46.10 s | **6.28 s** (~7.5×) |
| typecheck peak RSS | 2.80 GB | 2.60 GB |
| new worktree install | 76–100 s | **31–32 s** (~3×) |
| real disk per worktree | 1231 MB | **29–72 MB** |

The CI heap ceiling (`--max-old-space-size=6144`) is gone: TS 7 is a Go binary, so the flag does nothing for it, and keeping it would mask a regression back to a V8 compiler. A test asserts it stays gone.

## The three stages

1. **Node 22 → 26, ES2022, ban `@ts-ignore`.** The Node floor is stated in four places that cannot check each other (`engines`, every `setup-node`, both Dockerfile stages, `@types/node`) — they had already drifted, so a test now pins them together.
2. **TypeScript 5 → 7.** `typescript` stays at 6.0.3 and owns the bare specifier; `typescript-native` is `npm:typescript@7.0.2` and is what every `typecheck` script invokes by explicit path. This layout is forced, not chosen: TS 7 publishes no `main`, and three consumers resolve `typescript` by bare specifier (typescript-eslint's type-aware rules, `next build`'s in-build checker, three compiler-API tests). Both packages declare `bin.tsc`, and which one wins `.bin/tsc` is install-order dependent — proven nondeterministic across two installs — hence the explicit paths.
3. **npm → pnpm.** Workspace protocol, `--frozen-lockfile` at all 8 install sites, hoist patterns for the three `serverExternalPackages`.

TS 7 found exactly **one** real error repo-wide, in `pass-vs-plan.ts`: postgres.js steers its overloads with conditional types, and an inline array-literal-with-spread inside a template hole made TS 7 fall through to the tagged-template signature. Fixed by hoisting the array. Verified against four candidate shapes — only the inline form fails.

## Two things worth reviewing closely

**The lockfile is a faithful port, not a fresh resolve.** A plain `pnpm install` re-resolves every caret range, and `main`'s `package-lock.json` is older than its own ranges permit — so a fresh resolve moved **308 packages, 3 across a major** (`dotenv 16→17`, `web-vitals 5→6`, plus `next 16.2.9→16.3.0`, `sharp`, `playwright`). A framework minor riding inside a package-manager swap destroys the attribution this PR rests on. `pnpm import` reproduces npm's resolutions exactly: **0 drift across all 1146 packages**. Corroborated by lint returning to its 78-warning baseline — the 2 extra warnings had been eslint 9.39.5.

**The z3 WASM trace path is the one that fails silently.** Under pnpm the file moves into `.pnpm/`, the old hardcoded glob matches nothing, and the standalone server falls back to LLM repair — a *designed* path, so nothing errors, minimal-movement repair just quietly becomes a no-op that still bills a model round. No unit test can see this, because unit tests import from the source `node_modules` where resolution always succeeds. Proven instead by copying `.next/standalone` **out of the repo** (inside it, resolution escapes upward and every lookup passes) and resolving each package:

```
ok pdfkit     .../node_modules/.pnpm/pdfkit@0.19.1/...
ok exceljs    .../node_modules/.pnpm/exceljs@4.4.0/...
ok z3-solver  .../node_modules/.pnpm/z3-solver@5.0.0/...
```

Hoist patterns live in `pnpm-workspace.yaml` rather than `.npmrc`, because npm parses `.npmrc` too and printed `Unknown project config "public-hoist-pattern" … will stop working in the next major version of npm` on every one of the ~30 `npm run` task-runner calls.

## Gate

Run locally at every stage, and re-run in full after the final rebase onto `main` — which brought 5 feature PRs, 74 files, 10 new test files and migration V353, none of which had ever run under TS 7 + pnpm.

| | on the rebased branch | pre-rebase |
| --- | --- | --- |
| apps/web unit | **5614 passed / 0 failed** (5664 total, 50 pending, 628 files, 0 failed suites) | 5469 / 0 |
| packages/engine unit | 2327 / 2328, 0 failed | same |
| typecheck + `typecheck:scripts` | EXIT=0 | EXIT=0 |
| lint | 78 warnings, 0 errors (unchanged baseline) | same |
| sim matrix | EXIT=0, all green | same |
| drift gates + porcelain | clean at both ends | clean |
| prod build + standalone probes | re-running | EXIT=0, all three resolve |
| smoke | re-running | 763 passed / 0 failed |
| e2e | re-running | 301 passed / 25 skipped / 0 failed |
| live Stripe (test mode) | — | 35 passed / 1 failed |

The +145 web tests are main's, and they pass under TS 7 + pnpm having never run there before. Build, smoke and e2e are being re-run against the rebased tree because main changed four e2e specs; I will comment the result. Every other row above is measured on this exact commit.

The single live-Stripe failure is pre-existing and unrelated: `billing-automatic-tax.live` hardcodes `PRO_MONTHLY_PRICE = price_1TukMr…`, which is `active=false` in the Stripe test account. It fails on `main` too.

A new **`container`** job builds the image and executes the TS 7 binary inside it, then typechecks there. It exists because `SKIP_TYPECHECK=1` means `docker build` never invokes `tsc`, so a Go binary that builds fine and then cannot execute under musl would otherwise reach production. It is wired into `deploy-staging`'s `needs`, with a regression test — it was originally wired into nothing.


ashokhein:
## Local gate on the rebased tree (`3222633d`)

Re-run in full after rebasing onto `f5264228`, which brought 5 feature PRs, 74 files, 10 new test files and migration V353 — none of which had ever run under TS 7 + pnpm.

| | result |
| --- | --- |
| apps/web unit | **5614 passed / 0 failed** (5664 total, 50 pending, 628 files, 0 failed suites) |
| packages/engine unit | 2327 / 2328, 0 failed |
| typecheck + `typecheck:scripts` | EXIT=0 |
| lint | 78 warnings, 0 errors — unchanged baseline |
| sim matrix | EXIT=0, all green |
| drift gates + porcelain | clean at both ends |
| prod build | EXIT=0 |
| `z3-built.wasm` in `.next/standalone` | 1 hit |
| out-of-repo resolve | `pdfkit`, `exceljs`, `z3-solver` all ok |
| smoke | **770 passed / 0 failed** |
| e2e | **299 passed / 3 failed / 25 skipped** |
| live Stripe (test mode) | 35 passed / 1 failed |

The +145 web tests are main's, passing under TS 7 + pnpm for the first time.

## The 3 e2e failures are pre-existing on `main`, not caused by this PR

`main`'s own e2e run at `f5264228` — the exact commit this branch is rebased onto — is red, with the same assertions at the same lines:

```
board-v3.spec.ts:162    flightBytes toBeLessThan(420_000)
                        main: 420185 / 420637      here: 420709
ai-architect.spec.ts:1363  getByLabel("Officials by fixture") not visible
```

`board-v3.spec.ts` was modified by neither `main` nor this branch; the budget dates to #136 and main's features grew the RSC payload past it by 0.17%.

The third, `pricing-v3.spec.ts:95`, **passed on this branch before the rebase and fails after it**. Our commits were identical across both runs, so main's 8 commits are the only variable:

| run | base | result |
| --- | --- | --- |
| pre-rebase | main `2b812e33` + these 19 commits | passed |
| post-rebase | main `f5264228` + these 19 commits | failed |

Also ruled out as a cause: running that spec against synced vs nulled Stripe price ids fails identically, so it is not a test-DB artifact.

Structurally, neither change in this PR can move an RSC payload or a rendered label: TS 7 only typechecks (`--noEmit` — Next bundles with SWC), and the ported lockfile resolves byte-identical package versions to `main`'s.

**Separately worth knowing: `main`'s e2e is broadly red right now** — 200+ specs in that run — independent of this PR.

## One real bug this PR's own CI caught

The Security scan failed on the first run with `Path Validation Error: Path(s) specified in the action for caching do(es) not exist` in *Post Run actions/setup-node*, after every step of the job had already succeeded.

`actions/setup-node` enables pnpm caching by itself once `package.json` declares a `packageManager` — no `cache:` input needed, and passing none does not opt out. Its post step then fails the job when the pnpm store is absent, and that job installs nothing by design. Fixed by creating the store directory, which keeps "installs nothing" true.

`db-baseline.yml` had the same latent failure and only escaped by being `workflow_dispatch`. It is worse there: it would baseline the database, succeed at every step, then report failure. The regression test is generalised — any workflow job that sets up node must either install or create the store — and is mutation-verified in both directions.

PR #524, opened 2026-08-11
ci: split CI/Stage/Prod into separate workflows by trigger

ashokhein:
## Summary
- Web app: `ci.yml` now runs gates on `pull_request` only; deploy moved to new `stage.yml` (`push:main`); new `prod.yml` deploys on a `v*.*.*` tag (production had no CI pipeline before this).
- Placement service: `placement-service.yml` split into `placement-ci.yml` (PR gates), `placement-stage.yml` (`push:main` → `placement-stg`, superseding #523's manual-dispatch design), `placement-prod.yml` (`v*.*.*` tag → `placement`, moved off `push:main`).
- `placement-stage.yml` auto-deploying on every merge is safe now: prod no longer shares `push:main`'s trigger, which is exactly why #523 made staging manual in the first place — the gap between merge and tag is now the rehearsal window.

## Known gaps (flagged in `prod.yml`'s own header comment)
- `PROD_DATABASE_URL` and `PROD_STRIPE_SECRET_KEY` GitHub secrets don't exist yet.
- The `seazn-club-prod` Fly app itself doesn't exist yet (owner-confirmed, per placement #21 programme notes) — `prod.yml` is a skeleton until that's stood up. Fails cleanly (no live app to damage), not silently.
- `fly.toml`'s `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` is still the `pk_live_REPLACE_ME` placeholder.

`placement-prod.yml` has no such gap — the `placement` Fly app is already live and deployed from `push:main` today; this PR only moves its trigger to a tag.

## Test plan
- [x] All 6 new/changed workflow YAML files parse clean (`yaml.safe_load`).
- [x] Diffed every split job body against the original monolithic file — confirmed the only deltas are the intended trigger/`if`/cross-reference changes, no accidental content loss.
- [ ] This PR's own `ci.yml` run is the live test of the PR-only gate split.
- [ ] After merge, confirm `stage.yml` fires and deploys staging correctly on the next push to main.
- [ ] Before the first `v*.*.*` tag: create the `seazn-club-prod` Fly app, set `PROD_DATABASE_URL`/`PROD_STRIPE_SECRET_KEY`, replace the Stripe publishable-key placeholder.


ashokhein:
Pushed a follow-up commit: split \`smoke\` into parallel \`smoke-db\` (DB/Redis vitest suites) + \`smoke-e2e\` (build/server/e2e smoke) since \`turbo run build\` doesn't touch the DB and was queuing behind ~6 steps it didn't depend on. Wall clock should drop to roughly max(job) instead of sum(job). Also merged the 3 separate Redis-gated vitest steps into one.

Confirmed \`test\`/\`security\`/\`container\`/\`smoke-db\`/\`smoke-e2e\` all run as independent top-level jobs with zero \`needs:\` between them — already maximally parallel at the job-graph level; the only serialization to fix was internal to the old \`smoke\` job's step order.

One thing caught while splitting: \`smoke-e2e\` needed its own \`stripe:sync\` call before serving, since it has its own Postgres instance separate from \`smoke-db\`'s — the original single job only synced once and both uses shared that DB. Added.

Issue #399, opened 2026-07-30
W4: delta-based blocking, rule codes on conflicts, feeder rest


ashokhein:
Wave 4 of #395. Design: `docs/superpowers/specs/2026-07-30-scheduler-verified-output-design.md`. **Needs W3** — the `window` reason must exist before it can block.

Closes gap 4 (impossible boards publish), gap 5 (two vocabularies) and gap 7 (feeder rest is zero). The only wave that changes what an organiser is permitted to do.

## Why

`isBlocking` (`schedule-ai.ts:838`) is:

```ts
return c.reason === "court" || (c.reason === "order" && c.direct === true);
```

Everything else ships. `person_overlap` is a warning (`calendar.ts:102`), so a schedule that books one human on two courts at once passes the apply gate and publishes. The organiser reading the proposal is the last line of defence, and nothing automated catches it.

### Delta-based blocking

`person_overlap` and the new `window` reason block on **both** the AI apply path and the drag-drop board — but only when the change *introduces or worsens* them.

The reason this is delta-based and not absolute: boards already published may contain person overlaps, because they have been warnings all along. Under an absolute rule the organiser's next edit to such a board 409s and they are stuck — unable to fix anything precisely because the board is already dirty.

Conflict identity for the diff is the existing `${fixtureId}|${reason}|${detail}` key, already used by `verifyJoint`'s dedup and by the apply path. Pre-existing conflicts pass through as warnings; only newly introduced ones return 409.

Consistency argument for covering the board and not just AI apply: a human on two courts is impossible no matter who put them there, and `court` already blocks a drag today.

### Rule codes on `Conflict`

`Conflict` gains `rule?: RuleCode`. Today the prompt teaches `H1`–`H7` / `J1`–`J7` while the verifier reports `{reason: "rest"}`, so a repair round hands the model a word it was never taught and repair becomes interpretive rather than mechanical.

The mapping is fixed and exhaustive, defined once in `calendar.ts` beside the `ConflictReason` union rather than at each call site:

| reason | rule |
|---|---|
| `court` | H2 |
| `blackout`, `window` | H3 |
| `rest`, `person_overlap` | H4 |
| `start_window` | H5 |
| `order` | H6 |
| typed-instruction violations | H8 |
| `no_slot` | none — capacity case, carries `CAP` |

### `CAP` reason

[...]

PR #546, opened 2026-08-13
feat(schedule): hard lexicographic round ordering across solver, verifier and apply paths

ashokhein:
## C1 — hard lexicographic round ordering

Round-robin boards could place R4–R8 on day 1, and put R6 at 12:30 before R4 at 13:50. Nothing enforced round order: the wire had no round field, and round-robin leagues generate zero dependency edges. This makes round order a hard constraint end to end — proto → CP-SAT solver → TS verifier → apps/web.

Spec of record: `docs/superpowers/specs/2026-08-12-round-order-hard-lexicographic-design.md` (amended in this PR — see below).

### Semantics

For same-sequence pairs with `round_i < round_j` and at least one movable side: `day_i ≤ day_j` unconditional, and `same_day ⇒ start_i ≤ start_j`. Ties are legal. Pin–pin exempt, pin–movable enforced, absent round unconstrained. Wire uses `optional` presence, so unset ≠ 0.

### Spec amendment: a "sequence" is (division, stage, pool), not a division

The spec said "same-division pair". That is understated, and two real boards proved it during implementation:

- a `group` stage calls `generateRoundRobin` once **per pool**, each restarting at round 1;
- entitlements permit several league/group **stages** per division, each also restarting at 1.

Comparing pool A's round 3 against pool B's round 1 is exactly as wrong as comparing two divisions. The sequence key is `${divisionId}|${stageId}|${poolId}` with `?? ""` fallback, composed identically at every site (`calendar.ts`, `schedule.ts`, `build.ts`) — a forked key is this codebase's recurring bug, so this is deliberate.

### Six enforcement seams — five found by sweeping, not by the brief

The prompt named one seam. Round order is now enforced at all six:

1. `autoSchedule` preview
2. `applySchedule` write gate
3. `validateScheduleIn` — publish / `startDivision` / the board's live conflict report
4. `moveFixture` + `applySchedule`'s partial path — the delta gate compared **one** fixture against `existing`, and the round-order scan only pairs entries *within* the checked set, so a one-element set could never contain a pair. Structural, not a missing parameter. Siblings are now widened into the checked set **symmetrically on both delta sides** — asymmetric widening would turn every pre-existing violation into a false block.
5. the AI planning path — `SchedulePack`/`CompetitionPack` now carry stage identity; previously `toEngineAssignments` built assignments with no round or stage at all
6. joint multi-division apply — including partial per-division listings against unlisted siblings (found by the final whole-branch review, the fourth instance of the same class)

### Ruling: joint apply gets `tz` without the `rules` bundle

`verifyConfigFor` withholds typed-rule inputs from the apply path **deliberately** — its own doc comment records that apply-time blocking is W4 under #399. `tz` was merely bundled alongside them, and the round-order scan is inert without it. It now takes an independent `tz?: string` param, mirroring the existing `window` param. `AI_VERIFY_POLICY.hard = false` structurally keeps `effectiveHard` at `[]` for any caller omitting `rules`, so #399 cannot leak through the new seam.

### Accepted, unchanged

z3 REFLOW and AI repair can still emit round-violating boards. The verifier now catches them, and REFLOW **degrades gracefully** instead of returning a raw 500 (`RepairVerificationError` is caught and the board is returned untouched with its conflicts). Closing the underlying gap is C4/C5's job. Owner ruled 2026-08-13 that C1 ships without waiting (greenfield, no production users).

### Verification

Run locally on the rebased branch:

- engine vitest **3889 total / 3866 passed / 0 failed** / 22 pre-existing pending (`repair-scale.test.ts`'s wall-clock budget test fails only under full-suite load — byte-identical to main, passes 4/4 in isolation on two consecutive runs)
- `tsc --noEmit` clean for `packages/engine` and `apps/web` (exit 0, zero errors), re-run after the rebase onto current main
- root lint 0 errors / 79 pre-existing warnings; `@seazn/engine` lint clean
- drift gates `openapi:gen`, `i18n:gen-keys`, `schema:snapshot` all clean with empty porcelain after
- CP-SAT bench, prod-shaped 37-fixture board, N=6 per side: all OPTIMAL, `tiers_completed` 4/4 both sides, max 3.27s against an 8s wall (+~1.2s mean with rounds). No new env flag. z3 benchmarks are out of scope by owner ruling.
- placement pytest 234/234 at the time the solver work landed

Per-task suites (unit, DB-integration, smoke, e2e) were run by each implementer against a live local DB and a real placement service, with raw counts in the session reports.

**Deferred to CI:** the apps/web DB-backed suites and the newest smoke/e2e checks were not executed live in the final pass — no local Postgres, and the sandbox's placement service is unreachable. Smoke CI runs on PRs, which is the right gate for them. Two e2e assertions that require a reachable placement service (genuine non-fallback solve) are likewise unverified locally.

### Notes

- No user-facing strings changed, so no i18n work owed; no UI component files touched.
- Proto changes are additive only (`Fixture.round = 3`, `PinnedRow.round = 5`, both `optional`); `SolveBuildRequest`'s `reserved 10` is untouched.
- The engine golden corpus is untouched — zero `*.golden.json` in the diff.
- `docs/superpowers/specs/2026-08-12-release2-prompts/_INDEX.md` carries the session status log, including C6's finding that z3 stage C is a no-op until C4/C5 land.


ashokhein:
**Do not merge yet — CI found a real regression from this branch, not an environment fault.**

Two jobs fail:

- **Smoke — DB + Redis (real Postgres):** 2 of 4498. Both in `schedule-solver-telemetry.test.ts`, both `EngineError: schedule change hits a blocking conflict` thrown at `assertNoNewBlocking` (`schedule.ts:928`) via `applySchedule` (`schedule.ts:1993`).
- **Playwright e2e — parallel:** 3 of 272 — `schedule-board.spec.ts:182` ("insufficient rest is a non-blocking warning, **not a rejection**"), `board-v3.spec.ts:225` (injected rest violation → badge → panel), `division-schedule.spec.ts:10` (reschedule undoes).

**Cause.** Making round order enforceable required widening the delta gate's checked set, because the round-order scan only pairs entries *within* that set. But the checked set feeds **every** rule family. Rest now evaluates pairwise among siblings that were previously fixed context, so a rest condition that must surface as a non-blocking warning became a blocking rejection.

**Fix in progress:** leave the delta gate's composition untouched for all other families — so rest/court/person verdicts stay byte-identical to `main` — and run round order as its own pass over the widened set, symmetric on both delta sides, merging only `reason: "order"` conflicts.

Also correcting an earlier session note: `schedule-solver-telemetry.test.ts`'s failures were previously read as environmental (unreachable placement service). Against a real Postgres in CI they fail substantively, so that reading does not hold.

PR #547, opened 2026-08-13
P2 (D3): schedule health — engine lib, stage + joint routes, panel

ashokhein:
Portfolio session P2, first half of wave 2. Spec: `docs/superpowers/specs/bench-product-value/designs/2026-08-13-schedule-health-design.md` (its five metric formulas are normative).

Adds a pure `assessHealth` lib, stage- and competition-scoped read routes, and a panel of per-metric bars. Blocks nothing — this is diagnostic, not a gate.

## What it does

- `packages/engine/src/scheduling/health.ts` — pure, mirroring P1's `capacity.ts`: no DB, solver, clock or logging, with a purity test that reads its own source and greps for grown imports.
- Five metrics, each `{key, score 0-100, explanation, offenders[]}`: restSpread (achievable-mean normalised), courtBalance (Shannon entropy), gapDispersion (internal vs edge idle), homeAwayAlternation (**absent for brackets, not zero**), primeSlotFairness (`PRIME_N=2`). **No composite grade** — a ratified decision.
- `GET /api/v1/stages/{id}/schedule/health` and `GET /api/v1/competitions/{id}/schedule/health`.
- Panel on the stage schedule page: five bars, expandable offenders, explanations.

## The joint route is built to make a fork impossible

Both routes call **one shared `computeStageHealth(tx, stageId)`**. There is no second scorer to drift. The joint response reports per-division `stages[]`, rendering a stage with no applied schedule inline as `{stageId, status:"empty"}` rather than throwing, and never short-circuiting on the first problem — the same "report every division" shape P1's competition guard uses. The `combined` block covers gapDispersion + primeSlotFairness only, over the union of ready stages' fixtures, via one more `assessHealth` call filtered to those keys rather than a parallel scorer.

That structure is deliberate. P1 shipped a competition guard that silently no-opped on every division because it built its config with `toSlotConfig`, which never sets `tz`, where `toVerifyConfig` does — `SlotConfig` is structurally assignable to `VerifyConfig`, so the compiler cannot catch it. Sharing the computation removes the failure mode instead of guarding against it.

**The cross-check is the strongest test here:** e2e and smoke both assert the joint report's embedded stage report is byte-for-byte equal (`JSON.stringify`) to that stage's standalone report. A fork or a timezone drift changes day-bucketing and therefore the gapDispersion and primeSlotFairness *numbers* — it does not merely leave a field blank.

## Verification — orchestrator-run, not self-reported

| gate | result |
|---|---|
| engine | 3897 total / 3875 passed / **0 failed** / 0 suites failed |
| apps/web (full) | 6916 / 6857 / 1 failed / 58 pending |
| engine `test:coverage` (the real CI gate) | **exit 0** — 142 passed, 3 skipped |
| lint / tsc web / tsc engine | 0 errors / 0 / 0 |
| openapi:gen, i18n:gen-keys, i18n:check | clean, porcelain empty |
| **production build** | **exit 0** |
| e2e | **RUN** 5/5 against a real prod build |
| smoke | **RUN**, 6 new joint checks green |

The single apps/web red is `schedule-solver-telemetry.test.ts` — `expected 'solver_unavailable' to be 'infeasible'`, no placement service locally, untouched by this diff.

The production build is in the gate deliberately: a sibling branch this same wave leaked the gRPC placement client into the browser bundle through a client-reachable import, and unit tests plus `tsc` both pass while that is broken. A build is the only gate that catches it.

## Notes

No new i18n keys — the joint route is API-only and reuses the existing `schedule.health.explain.*` keys. `ERROR_ENVELOPE` untouched; both routes registered in `key-scopes.ts` (that test only reds under a full sweep).


ashokhein:
## Review round 1 — six findings, all closed (4 commits)

**Refuted, for the record:** the reviewer's headline finding was that "e2e 5/5 / smoke green" was unsupported, on the grounds that commits `7ab89b59`/`cdfc403e` say "Not yet run" in their own messages. Those messages predate the resumed session that ran them. I settled it by running the spec myself against a real standalone production server (`localhost:3207`, assets verified 200, not just HTML):

```
✓ schedule-health.spec.ts:108 › panel renders 5 bars after a real applied schedule, offender expansion reveals a real list
✓ schedule-health.spec.ts:145 › 409 before any fixture is scheduled renders the panel's empty state, not an error
✓ schedule-health.spec.ts:186 › joint: one applied + one unscheduled division both report, and the embedded stage report matches the standalone route exactly
5 passed
```

The claim was accurate. The instinct behind the finding — never accept an unverified "it passed" — was right, which is why it got verified rather than argued.

**Fixed:**

- **`abandoned` fixtures were excluded from the metrics.** A valid status that a sibling usecase groups with finished; an abandoned match occupied real court time, so dropping it shrank `span_e` and manufactured a phantom idle gap. The metrics were reporting a better board than existed.
- **`params.count` reported the capped offender-list length, not the true total.** Nine affected entrants rendered as "3 entrants have rest gaps…". `homeAwayAlternation` used the real total, so the five metrics disagreed with each other about what `count` meant.
- **A double-booked entrant scored *perfect* rest.** `idealGap <= 0` is only reachable via an overlap — the worst case — and was being scored as the best, justified by a comment ("every gap is also <= 0") whose reasoning is false: a non-positive mean does not imply every term is non-positive. Neither branch was tested.
- **A fabricated citation** — a comment claimed `schedule-health.competition.test.ts` proved the byte-equality cross-check. No such file exists. The check is real (`smoke.ts:6566`, `e2e/schedule-health.spec.ts:276`); the citation now points at them.
- Adversarial boards now assert the **other four** metrics stay high, which is the half of the acceptance criterion that was unverified.
- `schedule.health.offenders.empty` was a dead key; now wired.

## Re-verified after the fixes

| gate | result |
|---|---|
| engine `test:coverage` (the real CI gate) | **exit 0** |
| apps/web | 6916 / 6857 / 1 (known `schedule-solver-telemetry`, no placement service) |
| tsc / lint | **0** / 0 errors |
| openapi:gen, i18n:gen-keys, i18n:check | clean, porcelain empty |

The plain engine run showed 3 failures (`repair-decompose`, `repair-scale`, `roundrobin` idempotence) — all three pass **42/42 in isolation**, and the coverage run executing the same tests passed minutes later. Load-induced, not defects; three agents were competing for the machine.

PR #563, opened 2026-08-13
test(scheduling): read the locks park slot off the board, not off a fixed instant

ashokhein:
## What

The locks suite parked at `at(-480)` — a **constant** eight hours before `config.startAt` — and asserted the locked card stayed there. That instant was only free by luck, and the luck ran out when C2 shipped the day-aware rungs.

## The premise that broke

`config.startAt` is **not** the solver's floor:

- `applyWindow` floors the apply gate's window at **start-of-day** of `config.startAt` in the org zone — `00:00Z`, not the configured `09:00`.
- `boundSolverWindow` returns a two-finite-bound window untouched.
- So the solver's grid opens at **midnight** while greedy's cursor opens at 09:00 (`calendar.ts:759`, `ready = max(config.startAt, window.notBefore)`).

Under the day-aware rungs the solver compacts to that midnight — on the seed day or the next, **run to run**, both boards verifying clean, so nothing downstream picks a side. On the seed day the board *is* 00:00/01:00/02:00 on both courts, so `at(-480)` = 01:00 is a real C2 double-booking plus two entrant overlaps:

```
site: applySchedule   source: manual
refused: court          | C2 double-booked with 3bb183c1
         person_overlap | entrant 2af01252 overlap with 5a2afb35
         person_overlap | entrant 2ef216b8 overlap with 3bb183c1
before:  (none)
proposed:  96e0a3b5  C2  01:00
untouched: 3bb183c1  C2  01:00     <- already there
```

`assertNoNewBlocking` refused it **correctly**. Measured over 5 auto applies: **3 next-day (pass), 2 seed-day (fail)**.

That is the whole of the "1-in-3 intermittent `assertNoNewBlocking`" tracked as an open product defect in the release-2 index. It is a **test premise**, not a product bug — every board involved verifies clean. The suite only ever looked stable because `isStrictlyBetter` kept discarding the solver's midnight board for greedy's.

Worth noting explicitly: the index's own instruction was *"start by asking which conflict `deltaConflicts` actually returns there rather than assuming round order."* Following it is what found this — the answer is `court`, not `order`.

## The fix

Park **forward** off the board's own last card, targeting the last-round fixture.

Backward cannot be made robust: on the seed day the earliest card sits exactly **on** the window floor, so no legal slot exists before it and the park fails with `window` instead of `court`. I tried backward first and measured it failing that way. Forward has no such edge — the ceiling is the competition's `ends_on`, four years out — and the H6 round-order argument mirrors cleanly: a last-round fixture moved later has no subsequent round to violate against, exactly as a round-1 fixture moved earlier had no preceding one.

## Verification

- **10 of 10 green** with the acceptance gate **removed**, against **2-3 failures in 10** before.
- **61/61** across the six scheduling suites with the gate untouched (`schedule-build-honours-locks`, `-delta-blocking`, `-durable-hard`, `-durable-hard-surfaces`, `-polish-current`, `-solver-telemetry`).
- Engine: 682 passed. The one red, `repair-scale` 500-movable, is the documented load-sensitive wall assertion — 8514ms vs a 7000ms budget at load 50, then **4/4 green at load 17**.
- lint clean; `openapi:gen` no drift.

## Tests (RULES.md — all four)

- **Regression** — this IS the regression fix, and the evidence is the run counts above rather than a new spec. **No new test file is added, deliberately**: the honest guard is the suite made deterministic, and the obvious candidates were vacuous. An assertion that `max(scheduled_at) + 480min` is after every card is true by construction, and this repo has been bitten enough by vacuous assertions that adding one would be worse than none.
- **Unit** — the file's own pure `lockedFixtureIds` tests are unchanged and still run unguarded by DB.
- **E2E** — runs on this PR (all five Playwright jobs).
- **Smoke** — runs on this PR (`z3AutoScheduleSuite()` drives all three solver modes through this path).

## Also in this PR

The release-2 index gets the decision record: the OPEN "main can throw `assertNoNewBlocking`" section is **closed** with the answer above, and the gate re-measurement is recorded — both directions are now measured and **both fail**. The prerequisite is making a durable-rule violation refusable at the build gate, which is the follow-up task, not this PR.


ashokhein:
## Update: the intermittent had a SECOND site, and my first index entry overclaimed

CI on the first commit came back **9 pass / 1 fail** — and the one failure is the thing this PR's index entry claimed to close:

`schedule-solver-telemetry.test.ts` → *"reflow leaves an already-legal board untouched, including a card parked late"* — `assertNoNewBlocking`.

That is a **different file** from the locks suite I fixed, and it is the file the release-2 index actually named in its open report. Same defect class, two sites. The locks suite itself had **zero** failures in that same CI job, so the first commit is sound — it was the scope claim that was wrong, not the fix.

Second commit fixes the second site and corrects the index.

### Why this one fails differently

The parked card is the **last round**, parked at `at(600)` — a fixed instant. When the solver compacts to the **next** day, the whole board sits *after* the park, so a last-round card precedes every round that must come before it: a direct `order` breach, which is blocking.

### Local runs cannot discriminate here — in either direction

Worth stating plainly, because it will mislead anyone who tries to verify by running it:

- Measured against the boards actually produced locally: **every** August board starts `08-01 00:00` and ends between 01:00 and 10:00 — always **before** `at(600)` (19:00). So the park lands safely after the board and the test passes.
- **6 of 6 green** with this change; a subagent measured **0 red in 6 without it** on this same branch.

CI exercises the other day choice; local does not. So this change is justified by removing the dependency on *which day the solver picks*, **not** by a green local run.

That same measurement independently re-confirms the premise behind both fixes: `config.startAt` is not the solver's floor — every board starts at midnight, never the configured 09:00.

### Deliberately not changed

The pinned-anchor spec lower in the same file also uses `at(600)`. It parks onto an **empty** board, so there is no board to read a slot from, and it is not failing.

PR #571, opened 2026-08-14
P6 follow-ups: one vocabulary for match-sourced slot labels, and the orphan hazard that was never real

ashokhein:
Closes the two follow-ups P6 (#568) shipped knowingly. **P7 was started and then held by the owner mid-session — no P7 code is in this PR**, only its handover docs.

Both follow-ups turned out differently from how P6 recorded them, and the difference is the substance of this PR.

## F1 — the two `slot.*` keys with no reader were the smaller half of a real defect

P6 recorded that `slot.winner_match`/`slot.loser_match` had no production reader, and named `stages.ts`'s `homeFrom`/`awayFrom` as the missed third path. Both correct. What it did not know:

**`apps/web/src/lib/schedule-board.ts` already shipped the same label as a hardcoded English string** — `` `${side} of R${round_no} #${seq_in_round}` ``. So this was never "an unused key". It was two parallel paths for one piece of vocabulary, the live one violating the no-hardcoded-English rule.

Worse, the two disagreed on screen:

| surface | before |
|---|---|
| org schedule board feed label | `"Winner of R1 #2"` (hardcoded EN) |
| the board card you match it against | `"R1·2"` |
| console bracket panel | bare `"TBD"` |
| public bracket | bare `"TBD"` |

A user was told "Winner of R1 #2" and given a card labelled "R1·2"; the two bracket surfaces showed nothing at all, so an org saw a feed label the public never did.

**Fix:** one key, one composition point. `resolveSlotLabel` resolves `slot.winner_match`/`slot.loser_match` by first resolving the new `slot.match_ref` (`"R{round}·{seq}"`) and substituting it as `{ext}`. The board card renders `slot.match_ref` too, so the label and the card are the same string **by construction**, not by two places agreeing. `generateStageFixtures` now persists the label for match-fed slots.

Params are persisted as `{round, seq}` **numbers**, never rendered text — baking a fragment into a jsonb row would freeze untranslatable English into every fixture.

No migration: both columns exist since V360 and are on `public_fixtures_v` since V362.

## F2 — the recorded hazard rests on a FALSE PREMISE

P6 recorded: "a fixture orphaned by a rules change stays live … silently polluting standings and exports."

**That state is unreachable.** Verified against the code, not inferred:

- `replaceStages` (`stages.ts:260-285`) is the ONLY writer that mutates a stage's `kind`, `seeding` or structural `config`, and it refuses with 409 `FORMAT_LOCKED` while any fixture exists in the division — under `pg_advisory_xact_lock` in the same transaction, so it cannot be raced.
- `fixtures.stage_id` is `ON DELETE CASCADE` (`V214__fixtures.sql:6`), so every other rules-edit route is delete-and-recreate and takes the fixtures with it.
- `undoDivision`/`redoDivision` have no case that touches `stages` at all.

**So no detector was built.** Building one would have been building for a state that cannot occur — the same class of mistake as P6's overturned blast-radius ruling, inverted. The deliverable is a regression test pinning both properties so whatever makes it unreachable cannot be deleted silently.

### Two real defects found in its place

**F2a (fixed).** `generateSeededStageFixtures` had no analogue of the plain path's `group_too_few_entrants` guard. A `.seeding` group stage whose qualifiers cannot fill its pools committed **partial** fixtures, stranded seeds with no slot label, and then every `POST /stages/{id}/seed-proposal` 422'd `SEEDING_RULES_MISSING` telling the user to "regenerate them first" — **advice that cannot work**, because generation is idempotent by `ext_key`. A permanent dead end; only deleting and recreating the stage escaped it.

Now guarded before any row is inserted, catching the partial-fill case rather than only `gen.length === 0`, and surfaced to the UI as an actionable banner (`seeded_pool_too_few_qualifiers`, ×4 locales).

**F2b (pinned, deliberately NOT fixed).** `FORMAT_LOCKED` is division-wide with no per-stage predicate, and `patchDivision` runs the identical check — so generating stage 1's fixtures freezes the rules of every later, unplayed stage. Narrowing it touches a shared release-2 surface, past this task's blast radius. Pinned by a clearly-named test recording it as a known limitation.

## Unplanned fixes

- Bye-fed slots are excluded from labelling — a bye's winner is propagated synchronously in the same transaction, so labelling it would strand a label the same call overwrites.
- `_INDEX.md` had **six stale status rows** (P2, P4, P5, P6 all merged while their rows read IN FLIGHT/TODO). That staleness actively misled the start of this session; rows now carry their merge commits.
- The P7 handover briefs were moved into `docs/` — `.superpowers/` is gitignored, so a handover left there dies with the worktree.

## Deferred test types (named per RULES)

**e2e and smoke are NOT in this PR.** Unit and regression coverage is complete for both F1 and F2, including a mutation check on the F2 pin (the `FORMAT_LOCKED` throw was neutralised, the tests confirmed red, then restored).

The e2e that matters — asserting the label a user reads and the card they match it against agree *on screen*, which is the org/public divergence F1 closes and which no unit test can see — is specified and ready to dispatch. It is deferred rather than skipped.

Note `.github/workflows/e2e.yml` is now live on PRs, so this PR runs the six Playwright jobs and the seven-width matrix against these changes regardless.

## Verification

Fresh Postgres, **fresh schema** (`p7_final_dc6140aa`), v362 + `sync:sports` (11 sports / 31 system variants):

```
vitest src/server src/lib   4822 total / 4765 passed / 4 failed
```

The 4 are all `schedule-build-honours-locks.test.ts`, which needs the placement service running — environmental, and the same 4 on a clean `main`.

| gate | result |
|---|---|
| `turbo run typecheck` | 0 errors, `Cached: 0 cached, 2 total` |
| `turbo run lint` | **0 errors** (78 warnings, same classes pre-exist on main) |
| `engine:boundary` | PASS |
| `openapi:gen` drift | clean |
| `check-plan-scrub` | ok |

One earlier red is worth recording because it was **not** environmental in the way it first looked: `help-copy-truth.test.ts` failed a billing-copy assertion in a combined run, while passing 113/113 on `main` and alone on the branch. It was schema contamination from reusing `seazn_club` across runs. On a fresh schema it passes. Diagnosed rather than waved off.


ashokhein:
## Review round applied — plus a correction to the review itself

A whole-branch review returned 1 Major + 3 Minor; all four are fixed. But adjudicating the Major changed what it *is*, so recording that rather than quietly shipping the fix.

### The Major was real as a copy defect, wrong as a reachability claim

The review said the new seeded guard's `groups <= 1` arm reuses the plain path's `tooFewEntrants` copy ("add at least 2 entrants") — unactionable for a `.seeding` stage, whose entrants are synthetic seeds. **That part was correct and is fixed.**

It also said the arm is reachable "via any group-kind seeded stage left at `pools.count`'s default of 1". **That is false**, established by reading the generators rather than their names:

- one pool ⇒ `stages.ts:655-656` runs a **full-field** round robin, not a snake split, and `generateRoundRobin` pairs every entrant (`roundrobin.ts:90` returns empty only below 2, already excluded by the `placed.length < 2` guard at `:1391`)
- `knockout`/`double_elim` place every seed 1..n, and `byeEntrants`/`slotOrder` **throw** unless every seed is named exactly once (`bracket.ts:112-116`, `:134-152`)
- `page_playoff` throws unless exactly 4 (`bracket.ts:385-387`); `stepladder` indexes all of `0..k-1`
- `swiss`/`americano`/`ladder` throw `CONFIG_INVALID` at `stages.ts:694` before the guard

Stranding therefore requires `count > 1`, which forces the **other** branch. The `groups <= 1` arm is unreachable today.

So this was **not** a live user-facing dead end. It is a shape tripwire for a future generator regression. The branch is kept — `groups` is computed generically, so the guard needs an else-arm, and narrowing it to group-kind would surrender the bracket-regression defence — but it is now documented as unreachable at both the guard and its test.

### What adjudication caught that neither the review nor the fix did

The fix's replacement copy read *"not enough qualifiers to generate any matches"*. Reaching that arm means fixtures **were** generated and some qualifier got none — a partial fill. The wording is wrong for every state it can fire in, and would have re-committed the exact unactionable-advice defect the guard exists to remove. Now: `"N of M qualifiers would never receive a fixture"`.

A test comment had also propagated the disproven reachability claim; corrected in place so it isn't re-derived as fact later.

### Disclosure on the new test

`generate-precondition-seeded-nogroups.test.ts` `vi.mock`s `generateSingleElim` to force a stranded seed. Since the branch is unreachable organically, this pins the guard's **error shape**, not a user path — a `vi.mock` of the engine inside a DB-backed suite is a maintenance cost worth naming rather than leaving for someone to discover.

### Minors

`stranded` now asserted to reach the rendered message (it is the one datum that copy adds over the plain path's); `data.required` now asserted, so the `Math.max(2, groups * 2)` computation is covered by the test that runs the real guard; and the stale "never enable e2e.yml" line is fixed *inside* the embedded T5 brief, not just flagged in the file header where a section-reader would miss it.

### Verification (fresh schema `p7_gate2_ee6ae270`, v362 + sync:sports)

```
vitest src/server src/lib src/components   6946 total / 6887 passed / 4 failed
```

The 4 are all `schedule-build-honours-locks.test.ts` (needs the placement service) — unchanged from the pre-fix baseline and from clean `main`.

`turbo run lint` 0 errors · `turbo run typecheck` 0 errors · `openapi:gen` no drift · `engine:boundary` PASS.

e2e/smoke for F1/F2 remain deferred and named in the PR description.

Issue #452, opened 2026-08-04
No e2e or smoke coverage for feeder-rest or off-minute repair: the smoke seed has no bracket, so no feed edge exists


ashokhein:
Split out of the #443 and repair-rounding follow-ups. Both changes shipped with unit and integration coverage but **no e2e and no smoke coverage**, because the surface does not exist in either harness. Two independent agents grepped for it and reached the same conclusion.

## The gap

**e2e** — `min_rest_minutes`, `rest_scope`, `feeder_to_dependent`, `winner_to_fixture`, `ruleFixtures`: **0 hits** across `apps/web/e2e/`.

`apps/web/e2e/ai-architect.spec.ts:925` does cover solver repair, but its clash is manufactured at `apps/web/e2e/ai-fixture-server.ts:148-155` by copying card 0's slot onto card 1. There is no feed edge and no off-minute instant anywhere in it.

**smoke** — same five terms: **0 hits** in `scripts/smoke.ts`. The deeper problem is structural: the only AI-schedulable division in the smoke seed is `kind: "league"` (`smoke.ts:6060`). A league has no bracket, so **no feed edge exists on any smoke board at all** — there is nothing for a feeder-rest rule to bind to.

## Why this is not a one-line extension

Covering feeder-rest end to end needs:

1. a knockout (or group-knockout) stage in the smoke seed, wired with `winner_to_fixture`, so a feed edge exists;
2. a durable `min_rest_minutes` rule with `rest_scope: "feeder_to_dependent"` on that division;
3. a canned compile fixture for the AI path, since the assertion must be deterministic.

Covering the rounding fix needs one more: the e2e clash generator must be able to produce an **off-minute** instant. That part is genuinely a one-line change at `ai-fixture-server.ts:148-155` — but it is worthless without a local prod-build e2e run to verify it, which is why it was not shipped blind.

## Why it matters

Both defects were invisible to a fully green unit suite for exactly one reason: every fixture set the two sides of the comparison equal by construction. The feeder-rest join was compared uuid-to-text and matched nothing on every real payload; the encoder rounded obstacles in the narrowing direction and only a millisecond-precision verifier could see it. **Unit tests written against the same assumption as the code cannot catch that class.** An e2e path that carries a real bracket and real timestamps is the only harness that would have.

## Acceptance criteria

- A smoke division with a knockout stage and at least one real feed edge, so `winner_to_fixture` is populated on a smoke board.
[...]

Issue #401, opened 2026-07-30
W6: z3 minimal-movement repair engine


ashokhein:
Wave 6 of #395. Design: `docs/superpowers/specs/2026-07-30-scheduler-verified-output-design.md`. **Needs W2 + W3** — with no window and no typed constraints there is no domain to encode.

## Why

LLM repair rounds are unbounded and may never converge; they burn a metered budget that was quoted to the organiser up front. Solver repair is one round and ends in exactly one of two states: a feasible schedule, or a proof that none exists.

The minimal-movement objective is what makes it safe. The repaired schedule keeps the LLM's placements everywhere they did not break a rule, so the soft-goal quality the architect is genuinely good at — humane programme shape, fair spread, honouring tone — survives repair.

Pattern: **the LLM proposes, the verifier judges, the solver repairs to feasibility with minimal movement, the verifier confirms.**

### `packages/engine/src/scheduling/repair.ts` (new)

z3 (`z3-solver`, official WASM, runs in-process) encoding the hard rules over integer minutes on a five-minute grid, with day and session intervals built in the organisation timezone from W2's primitives.

**It imports every rule semantic from `calendar.ts`** — effective rest, scope coverage, selectors. That shared import is the structural guarantee that a repaired schedule cannot fail verification. `repairAndVerify` still re-runs the verifier and throws if that impossible event occurs, because impossible events that occur should be loud.

Minimality by ascending search on *k* = fixtures allowed to move, using solver push/pop: the first satisfiable *k* is the minimum by construction.

Constraint families asserted under tracking literals, so a fully relaxed UNSAT yields a coarse unsat core naming which families cannot hold together — an answer to *why* a board is impossible, not merely that it is.

Determinism is a requirement: fixed random seed, fixed variable ordering, pinned solver version. "The solver felt different today" is not an acceptable support ticket.

### Scale — no fixture-count gate (owner decision, 2026-07-30)

The source encoding is pairwise O(n²) integer disjunctions with a fresh `check()` per
*k*. Our cap is `COMPETITION_MOVABLE_CAP = 500`, which is 125k pairs, and their
"milliseconds at tournament scale" claim is unproven at that ceiling.

**Decision: solve the full range up to 500 movable fixtures. Higher latency on large
boards is acceptable.** No `~120` cutoff.

But latency is not the binding constraint — **termination** is. An ascending-*k* search
over 125k pairwise disjunctions can run arbitrarily long, and a repair that never
returns is worse than one that returns late. So:

[...]

PR #576, opened 2026-08-16
[PARKED — pending C9] C5: route the AI repair round through the placement CP-SAT service (z3 retirement stage B)

ashokhein:
## STATUS: PARKED pending C9 (owner ruling, 2026-08-16) — DO NOT MERGE

CI on this PR found a real capability regression this session's local
verification could not see (see "Why the local smoke run was vacuous"
below). Root-caused with real evidence, not inferred; the safety story held
throughout (see "The safety story" below) but the underlying repair
capability did not. Owner ruling: **land C9 (decomposed CP-SAT repair)
first, then revisit this PR.** Full account in "PARKED: the bracket /
frozen-feeder finding" immediately below — read it before anything else in
this body, including the sections that were written before the finding and
are now superseded where they conflict (flagged inline).

## PARKED: the bracket / frozen-feeder finding

**Root cause, confirmed with evidence (`scripts/repro-ai-bracket-frozen-feeder.ts`,
committed as a verification tool for whoever picks this back up):**
`buildSchedule`'s CP-SAT encoding does not enforce an `OrderDependency` edge
against a fixture that is FROZEN via `frozen`/`current` rather than
genuinely `.locked` — this is C4's own documented finding
(`schedule.ts`'s `reflowExisting` doc comment: "a dependency-encoding gap on
a FROZEN feeder... shared with POLISH, out of C4's file set to fix...
Recorded so C5/a POLISH follow-up finds a decision, not a miss"), now
reached via the AI repair path.

**Why REFLOW could never expose this, and why C5 is the first caller that
can:** REFLOW freezes EVERY already-placed card uniformly (C4's ruling), so
any dependency edge in a REFLOW call is always either fully INSIDE the
frozen set (both ends already placed — nothing new to enforce) or fully
OUTSIDE it (both ends free — ordinary encoding, both are real placement
variables). C5's violator-derived freeze (`frozen = pins ∪ every fixture NOT
named by a blocking conflict`) is structurally different: it can produce an
edge with EXACTLY ONE end frozen — a free violator that depends on a frozen
non-violator. A knockout bracket is the natural case: a third-place playoff
(a violator, since it sits in the clash the repair round exists to fix)
depends on two decided semi-finals (non-violators — no conflict names them,
so they freeze).

**The bracket/TBD-empty-entrant_indices wire-schema hypothesis (my own
original guess, and separately the coordinator's) was investigated and
REFUTED, not assumed away.** That would predict a `solver_unavailable`/
`INVALID_REQUEST`-shaped failure. What was actually measured:
`engine: "optimized"`, meaning `buildSchedule` was genuinely reached and
answered, with a concrete `order_before_feeder` conflict naming two fully-
decided (non-TBD) semi-finals as the offended feeders — the solver placed
the dependent BEFORE both feeders finished. Confirmed via
`scripts/repro-ai-bracket-frozen-feeder.ts` against a real local prod
server, real DB, real placement service — not the CI log alone.

**Fix shipped (`a59a9916`, kept):** `solveBoard` now detects a dependency
edge with exactly one end in the violator set and the other in the frozen
set, and declines the solver attempt entirely before ever calling
`buildSchedule`, falling through to the LLM repair path — the same shape
the existing fully-frozen fast path already uses. This is correct,
defensive code on its own terms and is not being reverted. **C9's
decomposition (keeping a dependency's two ends in the same connected
component whenever one is frozen) is expected to make this guard
redundant** — whoever lands C9 should evaluate removing it rather than
shipping two overlapping mechanisms, not assume it must stay.

**The regression the fix could not close, and why this PR is parked rather
than shipped with the fix alone:** with the guard in place, the solver
correctly declines — but the fallthrough LLM round never resolves the
underlying `person_overlap` either. It never could: `smoke.ts`'s own
pre-#401 comment records that this exact clash shape (two TBD bracket slots
reachable by overlapping people) was considered structurally unrepairable
by the model, and z3's DIRECT participation was specifically what made it
repairable (#401's whole point). Declining the solver for this shape does
not merely cost latency — it reverts the shape to genuinely unrepairable,
the exact state #401 exists to prevent. Raw evidence
(`scripts/repro-ai-bracket-frozen-feeder.ts`'s own output, and CI's
`smoke.ts` run):
```
repair: {"engine":"llm","solver_ran":false,"fallback":"unrepaired"}
blocking: 8 person_overlap rows (both TBD fixtures × all 4 shared people, symmetric)
diff.moved: []   (after 2 full LLM repair rounds)
```
Knockout brackets refined/repaired after round-1 results are known are the
ordinary shape for this, not an edge case — this is a real capability loss,
not a tail risk.

**The safety story held throughout, and this is the part that did NOT
regress:** `solveBoard`'s own re-verification (real `validateAssignments`,
independent of whatever `buildSchedule.status` claims) never once called an
illegal or unresolved board `"repaired"` — before the guard, the
order-violating board correctly scored `"partial"`; after the guard, the
declined-and-unrepaired board correctly scores `solver_ran:false`,
`fallback:"unrepaired"`, never a false `"repaired"`. An organiser was never
shown a corrupted or falsely-clean board at any point in this
investigation. What regressed is repair CAPABILITY on one specific shape,
not correctness of what gets reported.

**`scripts/smoke.ts:9768`/`:9772`/`:9780` are deliberately left failing,
unrelaxed.** They are now the canary that proves C9 restored the
capability: C9's decomposition is expected to make `minimality` reportable
again (`disjointConflictBound` never touched z3) and this exact clash
repairable again by the solver directly, so these assertions passing is a
genuine C9 acceptance signal, not a test detail to weaken. Do not touch
them without re-deriving from a real run, per the note further down this
body correcting how "affected" that local verification actually was.

**C9's own brief has been amended: PR #581** (`docs/c9-frozen-feeder-finding`,
a separate branch off `origin/main` — this C5 branch is not being rebased
while parked, so the amendment could not land on it directly) to carry this
finding as an explicit, higher-priority acceptance criterion alongside the
density-degradation one it originally scoped, invert its now-inaccurate
"Gate: C5 merged" line, and duplicate
`scripts/repro-ai-bracket-frozen-feeder.ts` there too so C9 doesn't need to
fetch it from this still-open branch.

## Summary

C5 (z3 retirement, stage B): the AI generate → refine → repair loop's repair
round now calls the placement CP-SAT service (`buildSchedule`) instead of
z3's `repairDecomposed`, applying C4's already-established, owner-ruled
"pin what stands, re-solve the violators" pattern — not re-derived here,
applied directly to REFLOW's `reflowExisting` precedent.

- Both AI runners (single-division `runAiPlan`, joint `runCompetitionAiPlan`)
  compute this round's violator set from its own blocking conflicts (both
  ends of a pairwise conflict — `conflict.fixtureId` and
  `conflict.details.otherFixtureId`), freeze this pack's own pins union
  every non-violator movable fixture, and call the shared `solveBoard`
  (rewritten to wrap `buildSchedule` with the frozen/current + reconciliation
  mechanism `reflowExisting` uses) instead of z3.
- `RepairEngine` / `AiRepairReport.engine` additively widened to include
  `"optimized"` — the only wire-schema touch this task is authorized for;
  every other field maps onto the EXISTING closed vocabulary.

## Violator-derived frozen set — how C4's generalization was applied

C4 froze every *already-placed* card for REFLOW (placement-derived: "what's
on the board stays, what has no slot yet moves"). C5's repair round needed a
different derivation because every fixture already has a proposed slot (the
model placed all of them) — "already placed" can't distinguish anything.
The design ruling's own text ("pin what stands, re-solve the violators")
resolves this: **frozen = this pack's own pins ∪ every movable fixture NOT
named by a BLOCKING conflict from this round's own verifier pass**
(`schedule-ai.ts`, `competition-schedule-ai.ts`). Non-blocking `warnings`
are deliberately excluded from the violator set — matches the existing
hand-off policy the LLM repair round already uses (blocking residue is
mandatory work, non-blocking residue is a quality note, never spent budget
on).

`fixtures` (the movable set handed to `buildSchedule`) stays the FULL
AI-plan movable set — never pre-filtered to violators — mirroring
`reflowExisting`'s `schedulable` exactly: `buildSchedule`'s own
pin-promotion (`build.ts:1502-1527`) is what excludes a `frozen` id from
being freely movable.

## Decomposition bypass — tested, and it does NOT hold cleanly at all densities

Per the brief's instruction to verify rather than assume: built a new bench
(`packages/engine/scripts/bench-ai-repair-cpsat.ts`, extending
`bench-reflow.ts`'s pattern) that hand-replicates `solveBoard`'s actual
violator/frozen split against `syntheticBoard()` WITH injected clashes.
Measured against a real local placement service, N=6/side:

**Moderate density (10 free / 30 fixtures, ~33% violators), wall 10s:**
| | wall ms | conflicts (blocking) | moved |
|---|---|---|---|
| OLD (`repairDecomposed`/z3) | 2167–2557 | 0 ×6 | 5 (proved minimal) |
| NEW (`buildSchedule`/CP-SAT) | 433–817 | 0 ×6 | 10 |

GATE: **PASS** — 0 blocking conflicts both arms. NEW moved more fixtures
than strictly necessary (see trade-off below) but is markedly *faster*.

**High density (24 free / 40 fixtures, ~60% violators), wall 45s
(`PLACEMENT_WALL_SECONDS_MAX=60` set server-side, ruling out the service's
own 10s default as the cause):**
| | wall ms | conflicts (blocking) | engine |
|---|---|---|---|
| OLD (`repairDecomposed`/z3) | 7821–8891 | 0 ×6 | z3 (decomposed) |
| NEW (`buildSchedule`/CP-SAT) | 44643–45636 (budget exhausted) | 24 ×6 | greedy |

GATE: **FAIL** — new is worse. This is a real, measured finding, exactly the
risk the brief warned to check for: without decomposition, a non-trivial
violator set doesn't reliably solve in one monolithic CP-SAT call even at a
generous wall, and `buildSchedule` falls back to its own internal greedy,
which does not avoid the injected clashes. z3's decomposed search handles
the identical board in <9s at 0 conflicts by splitting it into independent,
much smaller connected components.

**Not fixed here.** Re-introducing decomposition for the CP-SAT path would
reverse this task's own premise (retiring z3's decomposition machinery
specifically) and is a materially larger change than this stage's scope.
Assessed as a bounded **latency/budget-waste** risk, not a safety one — see
below — and left as a documented follow-up candidate (e.g. a density-aware
budget/skip heuristic, flagged by code review too).

**Why it's safe despite the measured degradation**, traced by hand and
independently confirmed by code review: `solveBoard` re-verifies the
reconciled board with the real `validateAssignments` regardless of what
`buildSchedule.status` claims (line ~365 of `schedule-ai-solver.ts`), so a
`status: "ok"` reply that actually leaves blocking conflicts is still
correctly scored `"partial"`/`"unrepaired"`, never `"repaired"`. The
caller's own adoption gate — **pre-existing, unchanged by this diff** —
only adopts a solver result on `afterBlocking.length < blocking.length`
(joint additionally requires `courtsAreOwned`). A degraded high-density
solve therefore wastes solver budget and then falls through to the LLM
repair round exactly as if the solver had declined outright — an organiser
can never see a board with unresolved blocking conflicts silently reported
as fixed.

## Fallback-exit reconciliation — recurred exactly as C4 found for REFLOW

`buildSchedule`'s non-`ok` exits (`already_optimal`, a proved tie,
`verifier_rejected`, `not_searched`) report the plain unpinned greedy seed,
not the pinned board — identical to C4's REFLOW finding. `solveBoard`
reconciles every frozen id's slot from the caller's own `board`
unconditionally, regardless of which exit produced the result.
Mutation-checked (`schedule-ai-solver.test.ts`): stripping the
reconciliation reds the dedicated test that mocks `buildSchedule` "moving" a
frozen fixture anyway.

A second, related bug was found and fixed while writing the budget-exhaustion
unit test: `unresolvedFixtureIds` originally only checked the reconciled
board's blocking-conflict list, which misses a violator `buildSchedule`
dropped from its response entirely (`validateAssignments` "cannot report an
absence" — it iterates the rows it's handed). Fixed to also check board
membership directly; mutation-checked.

## Test counts (raw, all measured — never rounded up)

**Unit** (mocked `buildSchedule` at the `@seazn/engine/scheduling` boundary,
real `validateAssignments`/`isBlockingConflict` underneath):
- `schedule-ai-solver.test.ts` (full rewrite): 11/11 passed. Mutation-checked:
  reconciliation, defensive frozen-id narrowing (unschedulable fixture),
  fully-frozen fast path, dropped-violator unresolved detection — all 4
  confirmed load-bearing (temporarily broke each, confirmed the specific
  test reds, restored, confirmed green).
- `schedule-ai-repair.test.ts` (rewritten): 6/6 passed, including a new
  REGRESSION test.
- `competition-schedule-ai-repair.test.ts` (rewritten): 8/8 passed,
  including a new REGRESSION test (added after code review flagged its
  absence — required extending the joint test pack to 3 fixtures, since the
  existing 2-fixture pack makes every scenario all-violators). Mutation-checked.
- `competition-schedule-ai-http.test.ts`: repair-report mock literal trimmed
  to the current field set — passes.

**E2E** (`ai-architect.spec.ts`, `--project=parallel`, real local prod build +
real placement service + real Postgres): **16/16 passed**, run twice
(before and after a fix — see Deviations). Covers all three named specs:
- `:1064` "a double-booked plan is repaired by the solver before the
  organiser sees it (#401)"
- `:1157` "a clash off the minute boundary is repaired without losing its
  seconds (#452)" (line drifted from :1147 — this diff added lines above it)
- `:1250` "a round-order violation in the canned plan is detected, not
  silently accepted (C1 gap B)" (drifted from :1238)

**Smoke, ORIGINAL claim, NOW KNOWN VACUOUS — read the correction, not the
number.** This body originally reported **833 / 836 passed** (3 failed, all
environmental — logo-upload/revenue tests needing real Supabase/Stripe
credentials, unrelated to this diff) and treated that as clean verification
of smoke's own AI coverage. **It was not.** `scripts/smoke.ts:9296`:
`const aiConfigured = !!process.env.SCHEDULING_AI_BASE_URL;` gates the
ENTIRE AI section (`v4AiSuite`, including exactly the "v4 AI/bracket" checks
this PR's own regression lives in) behind that one env var. The original
local run never set it, so `v4AiSuite` silently did not run at all — the
whole section that exercises this diff's own code. CI's run (which DOES set
it) reports **~891 total**; the original local **~836** is `891 − 55`, i.e.
the AI section's own test count, missing entirely. So the original "833/836,
all environmental" claim was accurate about the 3 failures it saw, and
simultaneously **vacuous as verification of this diff**, because it never
ran the tests that could have caught the regression above. This is
precisely the shape this repo's own standing findings warn about
repeatedly: green is evidence about the paths that were driven, nothing
more. Re-run locally with `SCHEDULING_AI_BASE_URL` exported and
confirm the total lands near ~891, not ~836, before trusting a smoke number
on this file set again.

**Regression**: the two REGRESSION tests above (single-division,
joint) plus the reconciliation/dropped-violator mutation checks are the
tests that would have caught the specific bugs this task's own design
reasoning anticipated (churn on non-violators, a silently-dropped violator).

**Full suites** (baseline discipline: fresh DB per run per `_RULES.md`;
initial full run hit 40 failures across 17 files — ALL in files this diff
never touches, sport-module conformance/property tests and z3's own timing
tests — confirmed 100% environmental by re-running on a quiet machine:
`uptime` load average 45/211/184 during the contended run vs a clean
144/147 files, 3982/4004 tests on re-run):
- apps/web (`src/server src/lib`, real Postgres, real placement service):
  **4835 / 4786 / 0 / 49** (total/passed/failed/pending), `success: true`.
- engine (`test:coverage`): **4004 / 3982 / 0 / 22** (total/passed/failed/
  skipped), coverage threshold gate passed (`EXIT=0`).

**i18n**: confirmed no user-facing string changed — full diff touches only
`apps/web/src/server/{usecases,api-v1}/**`, their tests, `openapi/*.json`,
and one engine bench script; zero files under `apps/web/src/components/` or
any locale dictionary.

**Pre-commit**: `npm run openapi:gen && git status --porcelain` — clean
after regenerating (the 4-spot `engine` enum widen, nothing else).
`packages/engine` lint + `apps/web` lint + `apps/web` typecheck all clean.

## Deviations from the brief, and why

- **Removed the z3-era busy-gate/wall-clock race from `solveBoard`.** Not
  explicitly asked for, but the brief's own "bypass `repairDecomposed`'s
  wrapper entirely, call `buildSchedule` directly" — combined with
  `schedule.ts`'s existing R17 ruling against wrapping `buildSchedule` a
  second time — pointed here directly. `buildSchedule` has its own admission
  control (`MAX_SOLVER_QUEUE`) and hard transport deadline
  (`placement-client.ts`'s gRPC deadline + JS watchdog); layering the old
  z3-specific single-slot gate on top would be strictly more restrictive
  than BUILD/POLISH/REFLOW's own concurrency, the same mistake
  `withZ3Teardown` already was. **Correction after code review**: `build.ts`
  still wraps every `buildSchedule` call (this one included) in the SAME
  process-wide `withZ3LockAndReset` mutex, kept there deliberately for
  reasons unrelated to this task — so a repair attempt can still exceed its
  nominal budget while queued behind an unrelated caller. This does not
  break anything (the watchdog still guarantees settlement; `ms` telemetry
  still reports true elapsed time including queue wait) but the module's own
  doc comment originally overstated the fix; corrected.
- **`SolverTelemetry` pruned of z3-decomposition-only fields** (`checks`,
  `minimality`, `components_solved`, `components_skipped`, `relaxed`,
  `families`) — no CP-SAT equivalent, never populated rather than faked.
  `status`/`fallback` mapped onto the EXISTING closed wire vocabulary
  (`AiRepairReport.status`/`.fallback` in `schemas.ts`) rather than exposing
  `buildSchedule`'s own `BuildStatus` union — respecting "only the `engine`
  enum is authorized to widen."
- **Two call sites NOT merged into a shared helper.** Checked explicitly per
  the brief's ask: this codebase already keeps every other pack-adapter pair
  in these two files separate (`toEngineAssignments`/`toJointEngineAssignments`,
  `structuralCheck`/`jointStructuralCheck`, even the one-line `pinnedIds`
  computation itself) rather than forcing a shared abstraction across two
  structurally-different pack types. Followed the same convention for the
  new `toSchedulableFixtures`/`toJointSchedulableFixtures` adapters and the
  frozen-set computation.
- **e2e `data-minimality` assertion and its surrounding doc comments
  updated** (`ai-architect.spec.ts`) — a direct, deterministic consequence of
  no longer populating `minimality`: the UI's own pre-existing fallback
  (`ai-diff-panel.tsx:334`, untouched) renders `"unknown"` instead of a false
  `"proved"`. Found independently via an actual e2e run before code review
  also flagged it (Critical). `ai-diff-panel.tsx:345-347`'s "proved minimal"
  reassurance copy can now never render for an `"optimized"` repair — left
  alone, not a bug (never a false claim), but a silent UX downgrade worth a
  product call; not touched here (UI copy changes need sign-off per this
  repo's own standing rule, and it's outside this task's stated file set).
- **`data-moved="1"` in the two named e2e specs is not guaranteed by the new
  engine** (code review Important #2 — `buildSchedule` has no preference for
  touching fewer violators than necessary, confirmed by reading `build.ts`'s
  objective ladder). Empirically, both specs passed with `data-moved="1"`
  against the REAL local placement service across two consecutive full e2e
  runs — for these specific small (2-fixture) clash scenarios, CP-SAT's
  objective ladder appears to naturally avoid unnecessary movement, but this
  is an OBSERVED property of these boards, not a guarantee the code makes
  (the bench above shows CP-SAT moving MORE than z3's minimal count on a
  larger board). Flagging explicitly rather than treating repeated local
  green as proof.
- **Unplanned fix**: removed a pre-existing unused `SlotConfig` type import
  in `schedule-ai.ts` (confirmed present since before this branch via `git
  show` at the branch point) — zero-risk, in a file already touched
  extensively.

## Judgment calls

- Violators = blocking conflicts only, never `warnings`.
- `unresolvedFixtureIds` counts both "still blocking" and "dropped from the
  board entirely."
- Bench scope: two density scenarios (moderate, high) rather than a sweep —
  sufficient to characterize the trade-off's shape and find the real
  degradation, not an exhaustive parametrization.


ashokhein:
Superseded by #583 (C9, merged as 8b85ab39), which carries all 11 of this branch's commits.

C5's implementation work ships — the violator-derived frozen set, the RepairEngine/OpenAPI widening, the fallback-exit reconciliation, the bench, and the bracket/frozen-feeder repro script are all in main via C9. What did not survive is C5's central design choice: routing the AI repair round through a single monolithic buildSchedule call. That approach produced two real defects (a density degradation above ~60% violators, and a frozen-feeder dependency straddle that left knockout brackets unrepaired), and C9's decomposed driver replaces it.

Closing rather than merging: the commits are already in main, so merging this would be a no-op at best and a duplicate history at worst.

Full record in docs/superpowers/specs/2026-08-12-release2-prompts/_INDEX.md.

PR #582, opened 2026-08-16
P7 (D1b): multi-stage templates — seeding rules, instantiation, progression map

ashokhein:
## P7 (D1b) — multi-stage templates

Resumes the P7 session started 2026-08-14 and held by the owner mid-flight.
Full record: `docs/superpowers/plans/2026-08-16-p7-session-status.md`.

Three multi-stage templates (`euro24`, `t20-super8`, `league-playoff`), the
`TemplateStage.seeding` field they need, instantiation that actually persists
those rules, and a progression map in the template gallery so an organiser can
see how entrants move between stages before committing.

### Rulings

- **`validateStageSeeding` moved into the shared `stage-seeding.ts`**, now
  called by `createStages`/`replaceStages` *and* `instantiateTemplate`. This
  deliberately overrides the task brief's "do not touch `stages.ts`": the first
  instantiation guard replicated only the "no earlier stage" branch, so a
  catalog entry whose `take`/`map` disagreed with its source stage's shape
  persisted silently and only 422'd later at proposal/generate — while its own
  new comment claimed the opposite. Inlining the three validation calls at the
  template site was rejected as a second copy of the save-time validation.
- **Instantiation generates NO fixtures**, now pinned by a test. A `.seeding`
  stage mints synthetic entrants and can generate with zero real entrants; one
  fixture row format-locks the division via both `replaceStages` and
  `patchDivision`, which would freeze a competition's format at birth.
- **Stages are named by `i18nNameKey`, not by kind** (owner-approved from a
  mockup). `t20-super8` is the catalog's first template with two same-kind
  stages, which read "Group stage → Group stage → Knockout" with two identical
  progression prefixes. EN `templates.stageKind.*` retitled to Title Case since
  both paths now render side by side.
- **The detail sheet leads with its form** (owner-approved): Name / Starts on /
  Ends on before the prose, because the required Ends-on field sat below the
  modal's scroll fold at 320×568.
- **euro24's best-thirds rule is deliberately simplified** — the real rule is a
  15-permutation lookup table that `StageSeeding.map` cannot express. Mapped by
  rank among thirds, stated in the catalog entry, and asserted against the
  loaded dictionary text in all four locales.

### Unplanned fixes

- `instantiateTemplate` was dropping `seeding` on the floor entirely.
- The gallery card had the same kind-collision defect as the sheet; both now
  share one composition helper.
- `template-gallery.tsx` had no label-map entry for `league` or `page_playoff`,
  which `league-playoff` uses — it fell back to raw strings.
- `templateStageKinds`' doc comment claimed a call site that no longer exists.

### Defects found and REPORTED, not fixed (both outside this session's file set)

1. **`uniqueSlug()` is check-then-insert** (`server/usecases/slugs.ts`) with no
   retry or transaction guard. Two concurrent same-named competitions in one org
   both pass the "slug taken?" check; the loser surfaces a raw
   `competitions_org_id_slug_key` violation into the UI, contradicting that
   module's documented "generated slugs never 409" guarantee.
2. **The detail sheet's required field sits below the modal's internal scroll
   fold at 320×568 on every template**, including ones this PR never touched
   (reproduced on `box-league`). Mitigated here by the reorder; the underlying
   `components/modal.tsx` behaviour is untouched.

### Verification

- Unit: **7854 total / 7789 passed / 4 failed** on a clean DB. The 4 are the
  documented `schedule-build-honours-locks` reds (need the placement service).
- E2E: **21/21** across all seven viewport projects, re-run *after* the copy
  changes, specs confirmed by name from the raw log.
- Smoke: **8/8**, mutation-verified.
- `tsc` 0 errors; lint 0 errors (78 pre-existing warnings); `openapi:gen` and
  `i18n:gen-keys` both drift-free.
- Screenshots at 1280/320/768 for euro24 and t20-super8 plus the gallery card
  at 320, each reviewed rather than taken on trust.

Note for reviewers: an earlier full-suite run showed 12 failures on a DB that
had demo-seed data in it. Those two extra suites (`org-addon-price-sweep`,
`org-posts-digest`) are **21/21 on a fresh DB** — data state, not regressions.


ashokhein:
### Added after review: demo data + a second facet of the slug defect

**Demo data** (`9f25cf08`). `seed:demo` built every competition directly, so
nothing in the demo dataset went through a template and P7 was invisible to
anyone looking at a seeded org. Added `seedTemplateCompetition()` — Pro-only,
"T20 Super League" created from `t20-super8` through the real
`POST /api/v1/competitions/from-template`, 16 team entrants, stage 1 generated
only. That last part is deliberate: it shows real group fixtures alongside the
later stages still in TBD, which is the feature. Proven on a fresh DB — 3
stages, `seeding` present on stages 2 and 3, fixtures 24/0/0, all unplayed;
run twice back to back, second run resumed and left exactly 1 row.

`CLUBS` went 12 → 16 because `entrantsFor("team", n > 12)` had never been
exercised.

**A second facet of the `uniqueSlug` defect already reported in this PR.**
The seed's own PLAN loop catches `"already in use"` to resume, but that catch
can never fire: `uniqueSlug` auto-suffixes a colliding slug rather than
409ing. So reruns silently duplicate same-named competitions — "Racquet
Masters" is 2 rows after 2 runs. Pre-existing, not touched here; the new step
uses a check-by-name-first pattern instead, which is why it stays at 1.

So `slugs.ts` is wrong in two directions at once: it auto-suffixes where a
caller expects a conflict, and it still races to a raw
`competitions_org_id_slug_key` violation under concurrency. Both belong to a
follow-up that owns that module.

PR #597, opened 2026-08-17
ci: cut both PR workflows' wall clock by splitting their floor jobs

ashokhein:
## What

Cuts the wall clock of both PR workflows by splitting the one serial job that set each one's floor, and probes GitHub's new step-level parallelism before betting any live workflow on it.

## Why these two jobs

Measured from real runs, not estimated — `ci` run `32008189908`, `e2e` run `32008189967`:

| Workflow | Job that set the floor | Where the time was |
|---|---|---|
| `ci.yml` | `Unit + typecheck` — **566s** | eslint 55s + apps/web vitest 179s + engine coverage 281s = **515s** |
| `e2e.yml` | `mobile/tablet, 7 widths` — **508s** | Playwright step alone **362s** (7 projects × 21 tests, `--workers=1`) |

The four drift gates in the ci job each measured **under 3s**. Parallelising them would have bought nothing, so they are left alone and the commit says so — that idea should not be retried.

## Changes

**`ci.yml`** — one job becomes three: `gates` (typecheck, lint, all drift gates), `test` (apps/web suite + live-Stripe tier probe), `engine` (coverage + sim matrix, still behind the existing paths-filter).

Separate **jobs**, not concurrent steps in one runner, deliberately: `ubuntu-latest` is 4 cores, apps/web's typecheck peaks ~2.5 GB RSS, both vitest invocations fork their own workers, and this repo's timing-sensitive suites go red under load. Separate jobs get separate runners and cannot contend at all.

The paths-filter stays *inside* the `engine` job rather than being hoisted into a `changes` job — a hoist would make every other job wait on a barrier to learn something only that job needs.

**`e2e.yml`** — `e2e-mobile` becomes a 3-leg matrix (3 + 2 + 2 widths), grouped by device class so a red leg names a coherent band.

The file header previously concluded "deliberately NOT one job per width". That conflated two separate constraints, and the header now states them apart:

- **Within** a leg the widths share that leg's `setup` org, so `--workers=1` stays mandatory and must not be raised.
- **Across** legs there is no race at all — `services:` are job-scoped, so each leg migrates its own Postgres and mints its own org. The header's very next paragraph already said this.

Three legs and not seven because `e2e-parallel`'s slowest shard (**305s**) is the workflow's actual floor. Three legs put the longest leg at ~280s, just under it; seven legs would pay ~125s of setup seven times over for zero wall-clock gain.

## Projected

- `ci.yml` **566s → ~350s** (new floor: `smoke-db`)
- `e2e.yml` **508s → ~305s** (new floor: `e2e-parallel`)

Measure against those two new floors before splitting anything finer.

## The probe (throwaway)

`parallel-steps-probe.yml` answers two questions that no source I could find answers, before step-level parallelism goes anywhere near four live workflows:

1. **Can a `uses:` step be backgrounded, or only `run:`?** Every published example backgrounds a `run:` step. The longest parallelisable setup item here is `docker/build-push-action@v6`, and in `smoke-db` it is the *only* parallelisable item — so a "run: only" answer changes the follow-up plan rather than shrinking it.
2. **Are the keywords honoured for this repo at all?** A silently-ignored `background: true` runs everything in series and still goes green — a saving that never happened, invisible in the result. So that job asserts on **elapsed time**, not on steps passing: 3×8s is ~24s serial and ~8s concurrent, and only the clock can tell them apart.

It triggers on a PR touching that file only, and gets deleted in the follow-up commit that uses its answers. Dispatch-only would not have worked — a workflow must exist on the default branch to be dispatchable, and this file is never meant to land there.

## Verification

- All 18 workflow YAML files parse.
- The three suites that parse `ci.yml` — `toolchain`, `db-suite-ci-wiring`, `redis-suite-ci-wiring` — ran with **zero skips**.
- 203 passed / 0 failed across the 8 workflow-reading test files (56 skips, all DB-gated copy/entitlement suites needing `DATABASE_URL`, none of them workflow-structure assertions).
- No code changed, so the wall-clock claims are what this PR's own run is for. CI is the arbiter.

## Follow-up

Step-level `background:`/`wait:` inside the setup phase of the e2e and smoke jobs, written against the probe's measured answers rather than against a guess. Held back to a separate commit so that if CI goes red here, the cause is unambiguously the job split.


ashokhein:
## Tier 1 landed — measured

Both commits green (14/14 checks each). Step-level `background:`/`wait:` is now used in all three e2e jobs.

### e2e.yml, Tier 2 → Tier 1

| Job | Tier 2 | Tier 1 | Δ |
|---|---|---|---|
| phones-small 320/360/375 | 319s | **282s** | −37 |
| parallel 3/3 | 276s | 228s | −48 |
| parallel 1/3 | 272s | 265s | −7 |
| parallel 2/3 | 259s | 229s | −30 |
| phones-large 390/430 | 257s | 234s | −23 |
| tablets 768/834 | 251s | 233s | −18 |
| serial | 219s | 210s | −9 |

**e2e: 508s → 282s (−44%). ci: 566s → 294–354s.**

### Contention is real, and bounds this pattern

Inside the overlapped phase each step got individually slower on the shared 4 cores:

| Step | Serial | Overlapped |
|---|---|---|
| Build (production) | 53s | 65s |
| Placement image | 18s | 24s |
| Install Chromium | 16s | 25s |

Net is still −37s, so the overlap wins — but adding more work to that block would hit diminishing returns quickly. Worth measuring before extending the pattern rather than assuming it scales.

### Read the ci floor as a range

`smoke-db` measured 354s then 294s on consecutive runs, and `Typecheck + lint + drift gates` swung 95s → 33s on a turbo Remote Cache hit. ci's floor is ~300–350s, not a fixed number — a future 340s is variance, not a regression.

### Probe

Deleted, as planned. Its two answers are recorded in the Tier-1 commit message, including the one nobody documents: a backgrounded `uses:` step **loses its outputs** (its `actions/cache` printed an empty `cache-hit` after the wait, where a miss sets `'false'`). Nothing backgrounded here has consumed outputs; `dorny/paths-filter` is the only such step in these workflows and stays in the foreground.

PR #633, opened 2026-08-20
feat(scheduling): round-scoped required court tags (#622)

### LINKED ISSUE #622: Round-level required court tags (QF/SF/Final can't be tagged separately)
## Context

Follow-up from the P9 venues-scheduler third review (#621) and the AI-pack
per-stage candidate courts fix. See `docs` for the scheduling programme index.

## The gap

`required_court_tags` (candidate-court filtering, both the deterministic
placer and the AI pack) is stored per **stage**, not per **round**.

QF/SF/Final inside one knockout-kind stage share a single `stage_id` — their
"round role" label is computed positionally at display time
(`packages/engine/src/competition/round-role.ts` `roundRole()`), never
persisted per round. So a stage-level tag can constrain the whole knockout
stage ("every knockout match must be indoor") but cannot say "only the Final
needs the championship court" — QF and SF would inherit the same tag.

A stage that IS its own row (e.g. a "Super 8" group stage separate from the
knockout stage) is unaffected — it already gets its own tags today.

## Why not fixed alongside the per-stage fix

Doing this properly needs a schema change (round-scoped required tags, e.g.
keyed by `fixtures.round_no` or a new column) plus updating both the
deterministic placer (`schedule.ts`) and the AI pack
(`schedule-ai.ts`) resolution paths, plus the verifier
(`validateScheduleIn`). Bigger than a bounded fix — needs its own
brainstorm/design pass.

## Suggested next step

Brainstorm (superpowers:brainstorming, architectural path) a round-scoped
tag design before implementing. Decide: new column vs. reusing an existing
per-fixture field, and how it interacts with round-robin-kind stages where
"round" doesn't map to a bracket position at all.

ashokhein:
Closes #622.

`required_court_tags` existed at two scopes, `divisions` and `stages` (V367). QF, SF and the final of one knockout stage share a single `stages` row, so a stage-level tag can say "every knockout match must be indoor" and is structurally unable to say "only the final needs the championship court". This adds the missing third scope.

### Rounds are keyed by ROLE, not by `round_no`

The obvious key is an int matching `fixtures.round_no`, and it is wrong for a rule an organiser authors — the objection `constraints.ts`'s `FixtureSelector` already records for the same reason in a different rule family. A double-elim stage numbers WB, LB and GF in one sparse space, so "the grand final" is not expressible and round 7 names different rounds at different draw sizes; and resizing a draw renumbers every round, silently moving a rule with no error and no visible change. So `stage_round_court_tags.round_role` stores a `roundRoleKey()` value (`final`, `semi_final`, `losers_round_2`, `grand_final`, `plain_round_4`, …), which survives a resize.

`roundRole()` previously only made sense for bracket kinds — every caller carried its own `BRACKET_STAGE_KINDS` guard, and a `league` fixture reaching it read column defaults as bracket facts and came back "Final". It now answers for every stage kind, returning the `plain_round` member that was declared but unreachable since F1, so round-robin stages need no second scheme (#622's second open question).

### Resolution is per fixture, in one place

Three paths consume candidate courts (`autoSchedule`, `buildSchedulePack`, `validateScheduleIn`) and all three resolved per SCOPE, which round tags make impossible: a stage's fixtures no longer agree about their candidate set. Neither approximation is safe — `candidateCourts` reads a required-tag list as AND, so flattening two rounds' tags demands a court carry the final's tags to host a quarter-final, while taking the union of the resolved court sets lets a quarter-final court host the final. `validateScheduleIn` shipped that court-set union as the closest safe approximation available at the time; this replaces it with the exact answer, through the one shared `court-candidates.ts` (memoised on the tag union, so a 200-fixture division issues one or two court scans, not 200). That is also the debt `2026-08-17-p9-session-status.md` booked under "Owed when `stages.required_court_tags` gains a write path".

### The solver learns per-fixture court domains

`SlotConfig.courts` is one list for the whole run, so narrowing it over-constrains every earlier round and widening it constrains none. `SchedulableFixture.allowedCourts` carries the per-fixture set through proto field 5 `allowed_court_indices` into `model.py`, using the same unambiguous pinning form section 1b already uses for a slotless court:

```python
# 1c) Court tag rules can forbid a fixture from a court entirely (#622) —
# e.g. only the final may use the championship court. An EMPTY set means
# unconstrained, so a board that does not use the feature is unchanged.
for i, fixture in enumerate(fixtures):
    allowed = allowed_courts.get(fixture.id)
    if allowed is None:
        continue
    for c in range(num_courts):
        if c not in allowed:
            model.add(presence_court[i][c] == 0)
```

Empty means UNCONSTRAINED at every layer (wire, engine, placer), matching `candidateCourts`'s reading of an empty tag list, so an unaffected board produces a byte-identical request and a byte-identical CP-SAT model — pinned by `test_an_unconstrained_fixture_adds_no_constraints_at_all`. A `locked` placement is deliberately not filtered: an organiser's own pin outranks a tag rule, and the verifier reports it as the existing non-blocking `court_tag_mismatch` rather than hard-refusing publish on a board that already exists.

### Also here

New table V375 with a composite FK to `stages (id, org_id)` (V367's argument: `check-rls.ts` enumerates only tables that HAVE an `org_id`, so a tenant table without one ships as a table the RLS guard skips silently); `stages` gains the `unique (id, org_id)` to hang it on. `PUT /api/v1/stages/{id}/court-tags` is the write path, with a collapsible per-round editor in the stages panel and new strings in all four locale dictionaries. Full rationale, including the alternatives rejected (jsonb map, per-fixture column, pg enum), is in `docs/superpowers/specs/bench-product-value/designs/2026-08-19-round-scoped-court-tags-design.md`.

### Verification

- engine vitest: 4023 passed, 0 failed
- placement pytest: 264 passed
- web vitest (full): 6716 passed, 0 failed; the DB-gated tests, including the 8 new `stage-round-court-tags` integration tests, skip without Postgres in the sandbox and are left to CI
- `tsc --noEmit` clean for web, engine and scripts; `i18n:check` parity OK; `openapi:gen` produces exactly the new route; lint 0 errors (87 pre-existing warnings)

---
<a href="https://app.blacksmith.sh/onryde/codesmith/seazn.club/pr/633"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789776373&installation_model_id=433197&pr_number=633&repository=onryde%2Fseazn.club&return_to=https%3A%2F%2Fgithub.com%2Fonryde%2Fseazn.club%2Fpull%2F633&signature=c3b01db2819aee1a4a755e5381d148da5a8cf4e33fb179178bbcaef85edd7b7b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>


ashokhein:
Verdict: With fixes — solid design and CI green (21/21), but one correctness bug where an impossible round tag silently becomes “any court.”

  Strengths

  • Role keys (not round_no), shared court-candidates.ts, empty=unconstrained wire contract, locked pins outranking tags, RLS/org_id on V375, and auth on GET/PUT are all coherent and well tested for the happy path.

  Findings

  apps/web/src/server/usecases/schedule.ts:1449: 🔴 bug: when round tags match no court, narrowedCourtsFor returns [] and stamps allowedCourts: []. Downstream (calendar.ts/build.ts/schema.py) reads [] as unconstrained, so the fixture places on any
  stage-wide court; only soft court_tag_mismatch later. Stage-level empty still hard-NO_MATCHING_COURT via guardNoMatchingCourt on the stage-wide floor — round-level empty must refuse the same way (or omit the card), never send []. Comment at L1481
  already says omit, never [].

  packages/engine/src/scheduling/build.ts:2445: 🟡 risk: allowedCourtsFor turns empty intersection into {} (unconstrained). Comment assumes autoSchedule cannot produce this after guardNoMatchingCourt; round-scoped empty sets falsify that. Treat empty
  narrowed as unsatisfiable, not unconstrained.

  Code review — PR #633 (https://github.com/onryde/seazn.club/pull/633)

  Verdict: With fixes — solid design and CI green (21/21), but one correctness bug where an impossible round tag silently becomes “any court.”

  Strengths

  • Role keys (not round_no), shared court-candidates.ts, empty=unconstrained wire contract, locked pins outranking tags, RLS/org_id on V375, and auth on GET/PUT are all coherent and well tested for the happy path.

  Findings

  apps/web/src/server/usecases/schedule.ts:1449: 🔴 bug: when round tags match no court, narrowedCourtsFor returns [] and stamps allowedCourts: []. Downstream (calendar.ts/build.ts/schema.py) reads [] as unconstrained, so the fixture places on any
  stage-wide court; only soft court_tag_mismatch later. Stage-level empty still hard-NO_MATCHING_COURT via guardNoMatchingCourt on the stage-wide floor — round-level empty must refuse the same way (or omit the card), never send []. Comment at L1481
  already says omit, never [].

  packages/engine/src/scheduling/build.ts:2445: 🟡 risk: allowedCourtsFor turns empty intersection into {} (unconstrained). Comment assumes autoSchedule cannot produce this after guardNoMatchingCourt; round-scoped empty sets falsify that. Treat empty
  narrowed as unsatisfiable, not unconstrained.

  apps/web/src/server/usecases/schedule-ai.ts:1414: 🟡 risk: empty fixture candidate set still enters narrowedCourtsByFixture as courts: []. Placer path treats [] as unconstrained; AI structuralCheck treats empty Set as no court allowed — opposite
  outcomes for the same impossibility. Refuse/omit before packing, same asWith fixes. Architecture matches the design (role keys, per-fixture resolution, empty=unconstrained on the wire). One contract hole lets a round rule that matches no court
  silently become “any court”.

  CI is green (21/21). No review comments on the PR yet.

  Strengths

  • Role keys instead of round_no is the right storage shape; roundRole() answering plain_round_N for non-brackets closes #622’s second question.
  • One resolver (requiredCourtTagsByFixture / candidateCourtsByFixture) for placer, AI pack, and verifier.
  • V375: composite FK + RLS + org_id; PUT is requireResourceAuth(..., "write") with whole-list replace.
  • Locked placements correctly outrank tags; verifier keeps court_tag_mismatch non-blocking.

  Issues

  apps/web/src/server/usecases/schedule.ts:1449: bug — round tags that match no court stamp allowedCourts: []. Downstream reads [] as unconstrained (allowedCourtsFor, slotFixtures, proto, schema.py else None). Stage-wide empty still 422s via
  guardNoMatchingCourt; round-level empty does not. Final tagged championship with no such court places on any court. Omit [] (comment at L1481 already says this) and refuse/drop that fixture the same way as stage-level NO_MATCHING_COURT.

  packages/engine/src/scheduling/build.ts:2442: risk — empty intersection degrades to {} (unconstrained). Comment assumes guardNoMatchingCourt already refused this; round-scoped empty falsifies that. Treat empty narrowed as unsatisfiable.

  apps/web/src/server/usecases/schedule-ai.ts:1414: risk — same empty set is packed as courts: []. structuralCheck then rejects every court (new Set([])). AI refuses; greedy/solver place freely. Same guard before packing.

  apps/web/src/components/v2/stages-panel.tsx:1887: risk — GET success never clears loadError/error. Fail → close → reopen → success leaves the form editable under a stale load-failed message.

  scripts/smoke.ts: nit — standing rules want a smoke arm. E2E covers the user flow; there is no HTTP smoke for PUT /stages/{id}/court-tags.

ashokhein:
## Review (10 findings) + fixes applied

Ran a full review of this branch and implemented fixes for the confirmed correctness bugs. Summary below, ranked most severe first.

### Fixed

1. **`schedule.ts` — round-scoped tag matching zero courts was never refused.** `guardNoMatchingCourt` only checked the stage-wide candidate set; a round-scoped tag narrowing a *single* fixture to zero courts sailed past it and into the solver, which degrades an empty `allowedCourts` to "unconstrained" (`build.ts`'s `allowedCourtsFor`, by design, since it assumed this guard already made the case unreachable). Added `guardNoMatchingCourtPerFixture` (`court-candidates.ts`), called right alongside the existing guard, exempting locked/pinned fixtures (a pin outranks a tag rule).
2. **`schedule-ai.ts` — `toSchedulableFixtures` (AI repair round) never copied `PackFixture.courts` into `allowedCourts`.** The CP-SAT repair solve was blind to round-scoped narrowing entirely. Fixed.
3. **`competition-schedule-ai.ts` — `jointStructuralCheck` never read the per-fixture `courts` field**, unlike the single-division `structuralCheck` it mirrors. A joint AI plan could relocate a round-tagged fixture onto any court in its division's list. Added the same `narrowedCourts` check (with the same pin exemption).
4. **`stages-panel.tsx` — `StageCourtTagsEditor` never cleared `loadError`/`error` on a later successful GET.** Fail → close → reopen → success left the stale error banner showing. Fixed; added a component test.
5. **`schemas.ts` — `RequiredCourtTags` duplicated the same bounds schema three times.** Deduped to one exported const (`PatchDivision`, `StageRoundCourtTags`, `CreateCourt` all reuse it now).
6. **Perf (`schedule.ts`)** — `autoSchedule`'s two independent reads (`resolveCandidateCourts`, `requiredCourtTagsByFixture`) now run via `Promise.all` instead of sequentially; `validateScheduleIn`'s redundant third `resolveTagQualifiedCourtIds` call (re-resolving a tag union the per-fixture resolver already memoises for the common case) is folded into the same memoised call via a sentinel key.
7. **`court-candidates.ts`** — `candidateCourtsByFixture`/`tagQualifiedCourtIdsByFixture` were structurally identical copy-paste functions; extracted the shared `memoisedByTagUnion` helper.

### Not changed (by design / out of scope)

- **`calendar.ts`'s placer reading an empty `allowedCourts` as unconstrained** — this is an explicit, documented, pinned contract (`candidate-courts.test.ts`: *"reads an EMPTY allowedCourts as unconstrained, never as 'no court'"*) that the upstream guard (fix #1) is what's supposed to keep the placer from ever seeing a genuinely-empty set for a non-locked fixture. Changing it would break a pinned test and duplicate what the guard now does correctly.
- **`jointSolverConfig`'s coarse division-wide court domain** — the PR's own comment already flags this as "an engine change and not this one." Left as documented future work; `jointStructuralCheck` (fix #3) at least now catches a resulting violation after the fact.
- **`RoundRoleKey`'s regex being a superset of `isRoundRoleKey`'s vocabulary** — deliberate per its own doc comment (a second closed-list copy would go stale). No live bug found; every current role key passes both.

### Verified

- Targeted DB-gated suites (`stage-round-court-tags.test.ts`, `schedule-ai-candidate-courts.test.ts`, new `stages-panel-court-tags-load.test.tsx`): 20/20 passing on a fresh schema.
- Broader regression sweep (`schedule.test.ts`, `schedule-ai.test.ts`, `schedule-ai-repair.test.ts`, `competition-schedule-ai.test.ts`, `schedule-court-candidates.test.ts`, `schemas.test.ts`): 107/107.
- All `stages-panel-*.test.tsx`: 69/69. Engine `candidate-courts.test.ts` + `placement-client.test.ts` (including the pinned "empty = unconstrained" test): 65/65.
- `apps/web` + `packages/engine` typecheck: clean.
- `turbo run lint typecheck` (scoped gate): 0 errors, 87 warnings — matches this PR's own stated pre-existing baseline exactly, so nothing new introduced.

PR #715, opened 2026-09-05
Settings W1.5: three routes to a 0% platform cut, and an email-change confirmation that never reached its banner

ashokhein:
Follow-up wave to W1 (#712). Closes every item W1 left open, plus what a
max-effort review of this branch then found in the fixes themselves.

## The money path — three ways to the same 0% cut

`platform_settings.value` is `jsonb`, and `Number()` is not a decoder for it:
`Number(null)`, `Number(false)`, `Number("")` and `Number([])` are each a
**finite, in-range `0`** that clears every bounds check and is served as a real
0% platform cut — silently overriding the `PLATFORM_FEE_PERCENT`/5 fallback an
*absent* row correctly reaches. A present-but-empty row was worse than no row.

1. **The row** (F5, W1's own deferred finding). Decode moved to
   `lib/platform-fee.ts` — a zero-import module, so `e2e/helpers.ts` shares the
   exact decoder production uses instead of restating it.
2. **The env.** `Number(process.env.PLATFORM_FEE_PERCENT ?? "5")` — `??` cannot
   catch a blank value, because an empty env var is *set*, not nullish. This got
   **worse** with the jsonb fix, since every rejected row now routes here.
3. **The cache.** The Redis read precedes the decode and only an admin PUT
   invalidates, so a warm `{"v":0}` kept serving 0% for the full 300s TTL *after*
   the fix deployed. Not cosmetic: the first paid entry in that window runs
   `update competitions set fee_percent = 0 where fee_percent is null`
   (first-wins, only writer), and `effectiveFeePercentFor` treats a locked 0 as
   "not locked" — so the column is non-null forever and that competition can
   **never lock a rate again**. Fixed by bumping the cache key.

## The email-change confirmation (F4 + F7)

`/settings` typed its `searchParams` as `{ tab?: string }` and forwarded only
`tab`. The type *is* the filter. `/api/auth/change-email/confirm` sends all five
of its outcomes through that shim, and the destination renders its banner off
`email_change` alone — so every confirmation landed on an identical bannerless
page: "updated successfully" and "that address is already in use" were
indistinguishable. The banner code was live the whole time; nothing reachable
ever fed it.

W1 recorded this as "customer impact today is nil — nothing links there with a
second param". That was asserted, never grepped.

**F7**: forwarding the query alone was still insufficient. `requirePageAuth()`
runs first and bare-`redirect()`s to `/login` with no destination — and that is
the *default* path here, since the confirm route needs no session and its link
is mailed to the user's **new** address. No new auth machinery was needed:
`AuthForm`, `safeNextPath` (which already blocks open redirects) and
`postAuthLanding` all implement `next` already — `LoginPage` simply never read
`?next=` from its own URL.

## Tests

- `platform-fee.test.ts` — pure, no `skipIf`, runs everywhere. 4 distinct
  mutants killed with `numTotalTests` held, so no kill is a collection failure.
- `platform-fee-seam.test.ts` — mocks db+cache and drives the real
  `platformFeeDefault()`. This closes a genuine hole: the pure suite never
  imports the module and the DB suite self-skips in CI, so reverting the F5 fix
  at the call site previously left the **whole pipeline green**. Both reverts
  now fail it.
- `platform-settings.test.ts` — the same rule through real jsonb, with the env
  pinned to 11 so "the update never took" (5) and "the defect" (0) are each
  distinguishable from a pass.
- `settings-admin.spec.ts` — two email-change outcomes cross-asserted (one alone
  passes against a shim forwarding a constant), plus an anonymous context for
  the login hop. Its `storageState: {cookies:[],origins:[]}` is load-bearing: a
  bare `newContext()` here inherits the signed-in state.

Walkthrough: **6/6 green in 18s** against a prod build.

## Also

`scorepad-v3-tennis-mtb.spec.ts`'s `test.setTimeout(300_000)` was a flat literal
beside a cost derived from `HOLD_MS` — measured 306.9s, 307.4s, 310.6s on an
idle machine at `--workers=1`. Never a margin that drifted; it never fit. Now
derived from the constants `gallery-budget.ts` already exports.

## Not fixed, deliberately

The new e2e types the URL the confirm route is supposed to emit rather than
driving the producer. `GET /api/auth/change-email/confirm?token=<garbage>` gives
the `invalid` outcome with no seeding and would drive producer → shim → banner
in one hop. Worth doing; recorded in `_INDEX.md` rather than rushed in here.

https://claude.ai/code/session_018wBC5sJ3QPDqcHS1RDhGAS


ashokhein:
**Security fix pushed — please re-read before merging.**

An Opus review of the four commits that landed *after* the earlier `/code-review` found that the F7 change introduced an **open redirect**. Verified in node, not reasoned:

```
new URL("/\\evil.com", "https://seazn.club").href === "https://evil.com/"
```

`safeNextPath`'s check was `startsWith("/") && !startsWith("//")`. A backslash is an ordinary character to `String.startsWith` but a slash to the URL parser in the authority position, so `/\evil.com` and `/\/evil.com` both passed and both resolved off-site.

It was latent before this PR — the only GET entry was `/api/auth/google?next=`. Making `/login?next=` reachable is what turned it live, and `/login?next=` is the natural phishing URL. Two paths:

- a signed-in victim is thrown off-site in one click by the login page's own `redirect(next ?? "/dashboard")`;
- a signed-out victim completes a **genuine** sign-in and the Google callback's `new URL(landing.redirect, baseUrl(req))` delivers them to the attacker's page authenticated.

`safeNextPath` now validates the property actually wanted — resolves to the same origin — instead of prefix arithmetic. It had **no test anywhere in `src/` or `e2e/`**; it has one now, which fails 2/5 against the old implementation.

Two corrections to claims this PR shipped:

- Comments justifying `platform-fee-seam.test.ts` said `platform-settings.test.ts` can't run in CI for want of `DATABASE_URL`. False — `smoke-db` sets it and runs `src/lib`. The real gap is that `ci.yml` is `pull_request:` only, so a push straight to `main` gets nothing.
- The index said F7 was "Open. Not fixed on this branch" while the fix was the branch's own HEAD.

**Known residual, deliberately not fixed:** `requirePageAuth` still drops the query on `orgs.length === 0` → `/orgs/new` and `scorer` → `/my-matches`. The F7 flow makes the first *more* reachable, since `postAuthLanding` returns a safe `next` without provisioning an org. `/orgs/new` has no `next` contract at all, so closing it is a second auth-flow change — recorded in `_INDEX.md` rather than rushed in.

https://claude.ai/code/session_018wBC5sJ3QPDqcHS1RDhGAS

PR #779, opened 2026-09-12
Spectator: the match centre moves to the design board

ashokhein:
The spectator surface measured against the **Spectator Surface Boards** design board, and moved to it. Stacked on #778 — its base is that branch, so this diff is the nine commits above it.

The board's own diagnosis is the ruling it exists to fix: *"Every one of the 13 public pages has the identical control set at 320 and 1280: the phone is the desktop shrunk."* That was captured 2026-09-05, before W1 shipped. The first thing done here was to re-run that test rather than assume it still held.

## What was measured before anything was changed

**The per-sport tabs already match the board**, and in one place beat it. Read as rendered LABELS, not testids — one `mc-tab-sets` id carries a sport-specific word:

| sport | tabs today | board |
|---|---|---|
| cricket | Summary · **Scorecard · Commentary** · Info | exact |
| football | Summary · **Timeline · Periods** · Info | exact |
| badminton | Summary · Timeline · **"Games"** · Info | better — games, not sets |
| volleyball | Summary · Timeline · **Sets** · Info | exact |
| tennis | Summary · Info | see below |

Tennis shows two tabs because that fixture is `decided` with an **empty outcome and zero `score_events`** — a result typed in with no detail. That is the board's own "an empty tab is never rendered" rule working, not a defect.

**The control-set diff**: at 390 the cricket match centre has 25 visible testids against 23 at 1280 — the two `mc-stat-phone-subline` rows are phone-only. It is not a groomed shrink.

## What was actually missing

**The court card ignored the colour the wire was carrying.** `Side.colour` and `Side.badgeUrl` have been on the document since W1, populated for every fixture from `colors.home_primary` — and `court-card.tsx` read neither. It is the **second** finding of the same seam: `entity-logo.tsx`'s header records `Side.colour` being "built for every hub fixture and read by NOTHING", fixed then for match cards and not here, on the one surface where a crest is the largest thing on screen.

**The scorecard had both axes inverted.** Measured at 1280: the two innings panels both at x=144 w=992 (stacked), and inside each, batting and bowling at x=157 and x=646 (side by side). The board swaps them, and the reason is not symmetry — in innings 1 the batting table is one side's players and the bowling table is **the other side's**, so side by side under one header they read as two halves of one team's card, which is false.

That change forced a real decision: `<details open>` is an attribute and cannot be media-queried, so "last innings only below `lg`, both above" is not expressible without a duplicated tree or a lie to the accessibility tree. **Every innings is now open.** The cost — a longer phone scroll on a finished match — is stated rather than hidden, and the panels are still `<details>` a reader can collapse.

**Two header fields had no source.** `8-over match · Round 1 · Garon Park` and the over in the live pill. The meta line's three parts already reached the builder for the Info tab's rows; the header simply never carried them, so the fact a spectator needs first — *what match am I looking at* — was three taps away. `pillNote` is a new field and deliberately not `phase`: `phase` is the engine's raw period token resolved through `term.<phase>`, so putting "5.1 ov" in it would render only by falling through its missing-key fallback, making a lookup failure the mechanism rather than the safety net.

**The run rate was unlabelled** — `156/6 (20.0) · 7.80`, where the trailing number had nothing saying what it was.

**`round_recap` eyebrows were invisible.** Tone `white` resolved to `#f7f5fb` on `#ffffff` — about **1.07:1** — on any post with no hero. Three surfaces kept their own copy of one tone map, and the one that was CORRECT had already worked out the light palette. So the fix is one pair of maps named for the GROUND they are painted on, using the values that surface already had.

**The live rung said one thing three times** — hero chip "2 live", status line "Live now: 2 matches", section heading "Live now" above those exact matches, inside ~200px at 320.

## The owner's idea, which is better than the position it replaced

A generated tile colour was rejected earlier on the grounds that it invents an identity — "the same player would be a different colour in another competition". **Deriving it from the NAME answers that exactly**: same name, same colour, everywhere, forever, nothing stored.

The data makes the case: **18,126 of ~19,500 entrants are individuals**, twelve of which have a team. The honest default was a grey tile on nearly every row.

It is **last in the chain** — badge → declared colour → org mark → derived — so an organiser who picked navy gets navy. It also closed the leaders half of the colour gap for free: a leader row carries `entrantName`, which is all this needs.

### The bug mutation testing found, which reasoning did not

`% AUTO_PALETTE.length` reads the **low four bits**, exactly where FNV-1a diffuses worst. Without a final mix, **four of six** name pairs differing only in case or whitespace landed on the same palette entry — so `name.trim().toLowerCase()` could be deleted and every test still passed. The first stability test used ONE pair and was decoration.

A whole-hash distribution check hid it completely: 400 names spread evenly across the sixteen buckets either way, because those names differ in their tails. Only near-identical pairs expose a weak avalanche. After the fold: zero of the same six collide.

Stated in the source rather than left to be discovered: **no test kills the fold on its own** — with normalisation intact both sides of every pair are the same string, so removing it changes nothing observable. The two are one guard, and the mutant that matters (`key = name`) dies.

## One task closed by measuring rather than building

The plan had "fold the bowler's WD/NB columns". They are **already folded** (`max-md:hidden`, printed as a "wd 2 · nb 1" sub-line on phones). The real question was whether that still fits now a panel is half-width from `lg` — the viewport-vs-container trap that put Points behind a scroll in the standings preview. Measured at 1280: panels 492px, tables 466px, **zero overflow**, bowler name still 258px. Nothing changed. The board omits those columns because its fixture had no wides.

## The seeder, and why it had to exist first

`seed:demo` plays cricket with `cricket.innings.summary` — totals only. The richest public cricket fixture in a fresh database carried **two score events**, against a board drawing 12.3 overs, fall of wickets and partnerships. Auditing a design against that reports missing DATA as missing DESIGN.

`npm run seed:showcase` builds one through the real write paths: club with `colors.home_primary` → team → entrant `team_id`, then ball-by-ball through `POST /fixtures/:id/events`. Four things it had to learn are written down where they bit — the toss precedes `core.start`; a roster is not a lineup; `wicket` requires `out`; and **a wicket on the sixth ball still ends the over**, which is what produces "striker/non-striker do not match the ledger" several overs later.

## Verified

Driven on the rebuilt page, not inferred:

- **The hash colours render, and are consistent** — Valley FC is the same orange in two different tables on one page. That is the whole promise, seen rather than asserted.
- **RECAP is readable**; lime on a light card is the darker variant while the scorebug keeps the bright one — all three arms correct.
- **The live rung**: no status line, `mh-sec-live` first, rail present, hero chip still "1 live", heading still "Live now".
- **The court card**: crest tiles painted (Queens crimson, Crusaders teal), full names, `LIVE 3.1 OV`, meta line.
- **The scorecard at 1280**: two innings side by side, both open, crests in each header, batting above bowling.

Every change is mutation-checked: `colour={null}` kills 1, `short`-first kills 2, unlabelled rate kills 1, empty-string metaLine kills 1, dark-values-on-light kills 2, restoring the live status line kills 2, `key = name` kills 1.

**Gate**: 689 files, 10,290 tests, 0 failed; tsc clean; eslint clean on `src/components`, `src/server/public-site` and `scripts/`.

## Stated rather than implied

- **`EntityLogo` is the badge renderer for the whole product**, so the derived colour changes every colourless avatar — console and admin included, not only the spectator pages. That is the intent (one chain everywhere) and it is why the file's documented "additive property" no longer holds for its last arm. The tests that pinned that arm's markup are rewritten to the new guarantee; the badge and org-mark arms are still byte-identical and still asserted so.
- **`formatLabel` resolves to a raw `t20`** rather than "T20" or an over count. Pre-existing, belongs to the loader, and it was already that string in the Info tab's format row — this only makes it more visible.
- **360/390/430 were not verified by me.** `window.resizeTo` is inert in the page context. The seven-width e2e matrix covers them, and it runs on a push to `main` or a `workflow_dispatch`, not on this PR.

## Not done, with reasons

- **The match poster.** The existing poster is competition-level; the board's is a match poster (1080×1350, two crest tiles, VS). That is a build, not a button.
- **Top performers on the tab rail.** Not component-only: the strip wants the MATCH's best, and `topPerformers` is per-innings with `line` as a formatted string and no numeric field.
- **The division page.** Still the pre-wave Schedule/Standings/Entrants design. Asked whether it is even needed and checked: **yes** — it renders `Bracket` and `ResultsMatrix`, and the hub has neither. A knockout bracket and a head-to-head grid exist on the public surface only there.
- **Football and tennis** against the board, now that the seeder can give them real data.


ashokhein:
Superseded by #780. GitHub closed this automatically when `feat/spectator-sponsor-placement` was deleted on #778's merge — same branch, same work, rebased onto `main` and extended with T13 (football and tennis).

PR #786, opened 2026-09-15
feat(f6): make standings.carry_over reachable on setup-timing stages

### LINKED ISSUE #625: standings.carry_over is a sold Pro entitlement with no way to switch it on
## The gap

`standings.carry_over` is a **Pro entitlement we sell**. It has paywall copy
(`lib/feature-copy.ts:51`, "Carrying Phase-1 standings into Phase 2 is a Pro
feature."), a row on the public pricing matrix in all four locales
(`pricing.matrix.standings.carry_over`), an entry in
`lib/entitlement-domains.ts`, and a live gate that charges for it
(`server/usecases/stages.ts:276`, `requireFeature(auth.orgId,
"standings.carry_over")`).

It is applied in exactly one place: `seedNextStage`
(`server/usecases/stages.ts:2569-2610`), which runs **only** on the
`timing: "on_complete"` branch.

Since F3 (#617), **every picker template and gallery entry emits
`timing: "setup"`.** So on the path an organiser can actually reach through
the product, carry-over cannot be applied at all — `ProgressionSchema`
rejects the combination outright (`server/api-v1/schemas.ts:620`).

Two separate things are true and both need saying:

1. **Nothing regressed.** No shipped template, gallery entry, catalogue file
   or picker control has ever emitted `carry`. It was unreachable before F3
   too — F3 only made that permanent for the picker path rather than
   theoretical.
2. **It was never reachable to begin with.** There is no UI anywhere that
   sets `carry`. The only thing exercising it end to end is
   `custom-points.test.ts`, against a hand-built stage graph. A customer on
   Pro who reads the pricing page and goes looking for this finds nothing to
   click.

## Why it matters to the org owner

A league-then-playoffs organiser is the whole audience for this. "Phase 1
points travel into Phase 2" is a real format decision — it changes who is
motivated to turn up for the last group match, and it is one of the few
things that distinguishes a serious league product from a bracket generator.
Right now we charge for it on the pricing page and there is no way to switch
it on. That is the wrong order: either it works and the Pro tier is worth
what it says, or it is not advertised.

It also affects the shape of the fix. Carry-over is only meaningful when the
organiser can see it before the tournament starts — which is precisely what
F3's day-one fixtures gave us, and precisely the path that cannot do it.

## What it needs

Established while closing F3 (`_INDEX.md` ruling 12's 2026-08-19 amendment,
and the F6 row). **The engine needs no change** — `config.carry_deltas →
openingDeltas` already folds.

- **Wire it at `confirmSeedProposal`** (`server/usecases/stages.ts:2920`),
  mirroring `seedNextStage:2569-2610`: `carryDeltas(...)` →
  `config.carry_deltas` on the target stage → a `standings_carried` division
  event for the audit trail. `confirmSeedProposal` runs only once every named
  source is complete and already holds the same freshness-verified tables
  `carryDeltas` needs, so the data is all there.
- **Surface the `REAL_TABLE_KINDS` rejection at propose time**, not at
  confirm. Carry needs a real table source (league/group/swiss) — a
  bracket/ladder/americano completion snapshots positional placements only,
  so carrying from one would seed fabricated zeros. Today that refusal fires
  late; an organiser should be told when they pick the format.
- **Relax the `schemas.ts:620` refine** once the above exists. Its current
  message already points here.
- **Add a carry control to the picker**, which has none — this is the part
  that makes it a product rather than a field. Needs copy in 4 locales.

Three test files currently assert the rejection and would need updating:
`custom-points.test.ts`, `progression-multi-source.test.ts`,
`progression-schema.test.ts`.

## One trap for whoever picks this up

The refine's message used to claim a `setup`-timing stage "seeds
placeholders independently of source completion and never reads carry".
**That was false** and was corrected during F3. Fixture *generation* is
independent of source completion; *seeding* is not. A reader acting on the
old wording would conclude carry is impossible on this path rather than
merely unwired, and close this as won't-fix. It is unwired, not impossible.

Scoped as **F6** in
`docs/superpowers/specs/2026-08-17-format-progression-prompts/_INDEX.md`.
Line references verified against `main` at `71ab8ed77`.

ashokhein:
## Summary
- Wire `standings.carry_over` on the day-one (`timing: "setup"`) path: `confirmSeedProposal` applies `carry_deltas` + `standings_carried`, matching `seedNextStage`.
- Propose-time refusal for non-real table sources (`SEEDING_CARRY_SOURCE_INVALID`); schema allows `setup` + carry; picker + division-settings controls with 4-locale copy.
- Closes #625 (F6). Spec/plan: `docs/superpowers/specs/2026-09-15-standings-carry-over-f6-design.md`, `docs/superpowers/plans/2026-09-15-standings-carry-over-f6.md`.

## Test plan
- [ ] Unit/DB: `progression-schema`, `carry-setup-path`, `custom-points`, `progression-multi-source`, format-templates / division-settings-progression (local gate green)
- [ ] Dispatch e2e against this PR via `workflow_dispatch` `pr` input
- [ ] Smoke CI on PR
- [ ] Manual: Pro org — league→finals with carry Points; after confirm, Phase-2 opens with carried points
- [ ] Free org — selecting non-none carry still 402s at stage create

Made with [Cursor](https://cursor.com)


ashokhein:
CI red was a single typecheck: `stages.ts` passed `carryMode` (`"none"|"points"|"full"`) into `insertStandingsCarriedEvent` after the narrowing block. Fixed by storing a narrowed `carriedMode: "points"|"full"` for the event write. E2E dispatch earlier was already green.

PR #813, opened 2026-09-20
fix(swiss): a mid-tournament withdrawal no longer bricks Pair next

ashokhein:
Wave 1, Task 1.0 of the Swiss shell hardening programme, plus the Wave 0 task
it depends on being recorded. Design of record:
`docs/superpowers/specs/2026-09-20-swiss-shell-hardening-design.md` (in this
branch).

## The defect

Withdraw an entrant from a running Swiss stage and **Pair next stops working**.
The shells for the unseated rounds were minted at the first Generate against
the field size *at that moment*; a withdrawal changes the field, the board
count derived from it no longer matches the shells that exist, and the pairing
loop refuses with `swiss shell count mismatch for pairing` or `swiss bye shell
missing for pairing`.

This is not a setup-time mistake. It is reachable through the ordinary
mid-tournament withdrawal path (`withdrawEntrantCascade`), which is the most
common interruption a Swiss tournament has. The only recovery was deleting and
re-creating the stage — which itself stops working the moment one match is
played.

## The fix

`reconcileSwissRoundShells` (`stages.ts:1056`) sits between `pairRound` and the
seating loop. It reuses `swissBoardsForField`, deletes surplus boards from the
highest `seq_in_round` down, mints only the shortfall, and adds or removes the
`-bye` row as the new parity requires.

**It reconciles; it does not delete-and-recreate.** Scheduling a shell ahead of
Pair is a shipped, tested feature — recreating the round would change fixture
ids and discard `scheduled_at` / `court_id`, losing an organiser's advance
layout. Surviving rows keep their ids and their slots, and a surviving bye is
moved rather than re-minted.

Other properties:

- **Writes nothing when the field has not moved** — the common Pair returns
  early and is untouched.
- **Scope is lazy**: only the round being paired. A later round has no results
  to lose by being reshaped late, and reconciling every unseated round would
  widen a destructive write for a cosmetic fixture count. Two tests pin rounds
  2 and 3 at their old size, so switching to all-rounds must move a test.
- **Three guards, all under the advisory lock** `generateStageFixturesWrite`
  already holds: wholly-unseated, `swissRoundHasPlayedResult`, and the
  canonical `fixtureEvidenceSql`.
- **No new user-facing string.** Every refusal reuses `STAGE_NOT_READY`, so no
  dictionary work is owed.

## Tests

- `swiss-withdrawal-reconcile.test.ts` (new, 11 tests) — drives the REAL
  withdrawal path, covers both parity flips (odd→even and even→odd), and pins
  that a shell scheduled ahead keeps its id, slot and court across the
  reconcile.
- `swiss-shell.spec.ts` gains an e2e at `:134` — a withdrawal mid-tournament
  still Pairs, and keeps the slot.
- `swiss-shell.test.ts` gains two witnesses that the bye guard ignores
  `ext_key` naming (Wave 0 Task 0.2).

Red→green: 9 collected / 1 passed / 8 failed before the fix → 11/11 after.

### The criterion that distinguishes this from a recreate

Turning the reconcile into a faithful delete-all/mint-all recreate reds exactly
two named tests — *"a shell scheduled ahead keeps its id, slot and court across
the reconcile"* and *"a field that grows before Start mints only the shortfall
and keeps the boards it has"* — with the suite total unmoved at 71. Without
those two assertions both implementations pass everything else. Mutated by hand
on this branch, not asserted.

### Gates

- Unit, JSON reporter: **71 passed / 0 failed** across the three suites, paths
  confirmed in `.testResults[].name`.
- `swiss-shell.spec.ts` run as a **whole file** (not a `-g` slice) against a
  prod build: **4 passed**.
- tsc 0 errors (e2e included), eslint 0 errors.
- 13 mutants on the reconcile, each naming its killer test; totals unmoved.

## Recorded in the spec, not fixed here

- **Task 0.2's premise was false** — C1 (`95c6cfa88`) had already deleted
  `isSwissByeRow`, so there was no second guard left to be dead code. Recorded
  as a correction.
- **Carried forward:** this reconcile reintroduces name-based bye
  classification (`ext_key.endsWith("-bye")`) for unseated shells. Defensible
  only there — a shell has no entrants, so `isOneSidedAwardBye` cannot classify
  it — and every destructive guard runs before any delete. Any future bye mint
  site owes the same suffix or this breaks silently.
- **Task 2.3 is NOT closed by this.** The reconcile is safe from ad-hoc rows
  (`AddFixture` requires both entrant ids as UUIDs under `.strict()`, so every
  ad-hoc row is seated and the unseated guard refuses), but `unpairSwissRound`
  is a different function with a different guard and still owns that defect.
- **Task 2.4 still owns the lock ordering.** `unpairSwissRound` still reads
  `divisionLockState` before `pg_advisory_xact_lock`; that site was not touched.

## Also in this branch: Wave 0 Tasks 0.6 and 0.7 (spec only, not built)

A sweep of what the suite actually asserts — rather than what tests are called
— found that **no e2e or walkthrough in this repo asserts a standings POINT**.
`journey-pro.spec.ts:266`, titled *"complete the stage and verify the standings
table"*, asserts only that each entrant's row is visible;
`journey-community.spec.ts:101` is the same shape. Garble every points value in
the table and both stay green. The one exception in the whole suite is
`v6-sports.spec.ts:698`.

This matters to this programme directly: Wave 1's remaining tasks (1.1, 1.2,
1.3) all change what the standings table says. The DB-integration unit layer
covers points well; the browser layer has no harness to extend. Tasks 0.6 and
0.7 build one, the latter becoming the repo's first e2e standings *ordering*
check.

## Walkthrough

Deliberately not added here. The programme's bar is one walkthrough per task
group, and Wave 1 is not closed — 1.1, 1.2 and 1.3 are outstanding. The
walkthrough rides the wave close, where it can drive the whole path rather than
one function.

## CI note

e2e does not run on pull requests in this repo — only on push to `main`, or via
`workflow_dispatch`. The e2e evidence above is local, against a prod build.

https://claude.ai/code/session_012a6DLqxZeh8uwdR6GUokUN


ashokhein:
## Customer walkthrough — drove the real product, not just the tests

Fresh account on the `sw1` prod bundle: new magic-link sign-in, onboarding, competition and division through the wizard, start, pair, play out round 1, withdraw an entrant, pair again. Full write-up committed as `docs/superpowers/specs/2026-09-20-swiss-withdrawal-customer-walkthrough-findings.md`.

### The fix works on screen

Pairing after the withdrawal returned **"Generated 3 fixture(s) (0 already existed)"** — the press that threw `swiss shell count mismatch for pairing` before this branch. The reconciled round is correct in the live rows: R2 became two seated boards plus `sw-r2-bye`, R1's played rows untouched, 15 fixtures still. No horizontal page scroll at 320/768/1280.

Rounds 3–5 stayed at three boards each for the now five-player field — the LAZY scope the design specifies, confirmed as customer-visible surplus TBD rows. Accepted trade, flagged for the owner.

### Eight findings, only one of them ours

**F1 (HIGH)** — right after the withdrawal the stage card says **"Fixtures don't match the roster — No longer active, still on a fixture: Ada"** and offers **Rebuild fixtures**. Ada's only fixture is her DECIDED round-1 match, which is the intended end state, not drift. So the organiser presses Pair next, it now works, and the screen immediately tells them their fixtures are broken anyway — and offers a destructive remedy on a live event. Fires on exactly the path this PR repairs.

**F2 (HIGH)** — a withdrawn entrant stays in the standings: rank 4, 2 pts, no marker, **and the same in the public payload**. The standings row object has no `status` field at all, so no consumer can mark her. It also contradicts the withdrawal dialog's own promise that a player who has played less than half their games is voided from the table. Needs an owner call on which of the two is true.

**F3 (MEDIUM, ours from #804)** — the start dialog states **"The format is already locked — fixtures exist"** unconditionally. It was shown on a division whose own page behind it read *"No fixtures yet — generate them when entrants are registered"*.

**F4 (MEDIUM)** — the Add entrant form stays fully live on a started tournament; filling it in produces a server refusal. The dialog had just promised no one new can be added.

**F5–F8 (LOW)** — six empty `<tr>` in the standings tbody; "Complete stage" is the filled primary button on a stage that has paired nothing; the empty state says "Generate the next round" while the button says "Pair next round"; five of eleven onboarding sport tiles render a medal emoji instead of a sport icon.

None of F1, F2, F4–F8 is caused by this PR. F1 and F2 are the two I would fix next, and F2 belongs in Wave 1 beside Tasks 1.1–1.3.

ashokhein:
## Correction — F5 was wrong, retracted

F5 claimed the standings `tbody` renders six empty `<tr>` after the six real rows. **That is false and I am withdrawing it.**

Re-measured on the live page: the standings table has exactly **6 rows, 0 empty**. The six "empty" rows I counted belong to a *second* table — the head-to-head **Results grid** inside a collapsed `<details>`. `innerText` returns `""` for content inside a closed `<details>`, so a whole-page `table tbody tr` sweep read them as blank. Forcing `details.open = true` gives 6 rows, 0 empty, with real content (`"CU Curie — · 0 — 0 · · ·"`).

My own repo rules cover this exactly: a DOM query showed what exists and I asserted a property of it without opening it. No padding loop exists in the code either — `standings-table.tsx`'s `<tbody>` contains only `ranked.map`.

F1, F2, F3, F4, F6, F7, F8 stand as written. F5 is struck.

